PR #10, opened 2024-03-14
Add multipass V0 (check for false-positives from matches from excluded paths)

fuchsnj:
This uses the new scope changes so that regex matches from excluded scopes are saved in a set, and if any match from an included scope matches, its considered a false positive.

The `logs-backend` has some code to prevent an excessive number of matches from being stored in the "excluded matches" set by only keeping the largest string for each rule / start offset. This is not needed here since the `regex` crate does not produce overlapping matches for the same rule (you will never have multiple matches from the same rule with the same start offset). 

The `logs-backend` also processes excluded matches in an order-dependent way. The "excluded matches" set is populated as fields are scanned, so matches are only considered a false-positive if the excluded match was scanned before. This is somewhat improved by having all the "custom" fields scanned before the "message" field, since most exclusions happen in custom fields. Within the "custom" / "message" fields though, the order is not well defined since it uses a `HashMap`.

In this implementation, the excluded matches are processed in an order-independent way. All of the excluded matches are collected, and then a 2nd pass is used to find all of the false-positives. In practice this should closely match what the `logs-backend` is doing, but there might be some differences. If the differences are large enough, I might add an additional PR to temporarily make the behavior more similar.


sds/src/scanner/mod.rs
```diff
@@ -96,113 +97,87 @@ impl Scanner {
...
         let mut output_rule_matches = vec![];
 
-        let mut rule_matches_iter = rule_matches.into_iter().peekable();
+        for rule_matches in &mut rule_matches_list {
+            // All rule matches in each inner list are for a single path, so they can be processed independently.
 
-        while rule_matches_iter.peek().is_some() {
-            let mut rule_match = rule_matches_iter.next().unwrap();
+            // TODO: remove the clone here
+            let path = rule_matches.first().unwrap().path.clone();
```
vinckama, line 121: This the path is extracted from the first match, won't it be better to remove it from the `InternalRuleMatch` and store it as a key of the `rule_matches_list`?

You won't need the clone anymore

vinckama (commented):
I've added a question, a suggestion and a thank-you.

sds/src/scanner/mod.rs
fuchsnj: Good suggestion, changed in https://github.com/DataDog/dd-sensitive-data-scanner/pull/10/commits/2534554df734ef16e737b27c9b8fb080f668a8d4

Note that while I added another `clone`, it replaces an `.into_static()` which was basically a clone.

sds/src/scanner/mod.rs
```diff
@@ -333,13 +305,14 @@ impl<'a, E: Encoding> ContentVisitor<'a> for ScannerContentVisitor<'a, E> {
...
             // so those don't need to be filtered out here
-
             let mut it = regex_automata::util::iter::Searcher::new(content.into());
             while let Some(regex_match) =
                 it.advance(|input| Ok(rule.regex.search_with(cache, input)))
             {
                 if exclusion_check.is_excluded(rule_index) {
-                    // TODO: Multipass V0 logic will go here.
+                    // Matches from excluded paths are saved and used to treat additional equal matches as false positives
+                    self.excluded_matches
```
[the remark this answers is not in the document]
fuchsnj: Fixed in https://github.com/DataDog/dd-sensitive-data-scanner/pull/10/commits/837c506796c55c928f404e63210d503805e156a2

vinckama (approved):

PR #11, opened 2024-03-14
Add github checksum as secondary validation

trosenblatt:
Add github checksum validation as secondary validation
[JIRA_LINK](https://datadoghq.atlassian.net/browse/SDS-197)

It adds 2 deps
- [crc32fast](https://crates.io/crates/crc32fast)
- [base62](https://crates.io/crates/base62)


sds/src/secondary_validation.rs
```diff
@@ -92,6 +96,32 @@ impl Validator for ChineseIdChecksum {
...
 }
 
+impl Validator for GithubTokenChecksum {
+    fn is_valid_match(&self, regex_match: &str) -> bool {
+        // Implementation of https://github.blog/2021-04-05-behind-githubs-new-authentication-token-formats/
+        let parts: Vec<&str> = regex_match.split('_').collect();
+        if parts.len() < 2 {
+            return false;
+        }
+        let last_part = parts[parts.len() - 1];
```
fuchsnj, line 106: parts.last() is more idiomatic here (It returns Option<&str>)
```suggestion
        let last_part = parts.last().unwrap();
```

fuchsnj (approved):
1 small nit

PR #14, opened 2024-03-15
sds-go: fix the go.mod pkg declaration and moved sds.h relatively to the CGo usage.

remeh:


sds-go/go/go.mod
```diff
@@ -1,3 +1,3 @@
-module sds
+module github.com/DataDog/dd-sensitive-data-scanner/sds-go/go
```
trosenblatt, line 1: @remeh just for my understanding what does it change for external repo that would use this ?

trosenblatt (approved):
LGTM with comment for my understanding

sds-go/go/go.mod
trosenblatt: Yes I got that but what does it change for outside if you let module sds instead ?

remeh: It can't be imported/resolved as `sds` is not the full address of the pkg.

A random example of another pkg that is resolvable from the outside, from another repo if that helps: https://github.com/DataDog/datadog-agent/blob/main/pkg/tagger/go.mod

trosenblatt: ok thks

PR #16, opened 2024-03-25
Add labels in RuleConfig

vinckama:
Some labels can be specific to the rule itself (`rule_id` for instance), the goal of this PR is to attach labels to `RuleConfig`. [ticket](https://datadoghq.atlassian.net/browse/SDS-123)

### DIFF
--- a/sds/src/rule.rs
+++ b/sds/src/rule.rs
@@ -4,6 +4,7 @@ use crate::match_action::MatchAction;
 use crate::path::Path;
 use serde_with::{serde_as, DefaultOnNull};
 
+#[serde_as]
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 pub struct RuleConfig {
     pub pattern: String,
@@ -12,6 +13,9 @@ pub struct RuleConfig {
     pub scope: Scope,
     pub proximity_keywords: Option<ProximityKeywordsConfig>,
     pub validator: Option<SecondaryValidator>,
+    #[serde_as(deserialize_as = "DefaultOnNull")]
+    #[serde(default)]
+    pub labels: Vec<(String, String)>,
 }
 
 impl RuleConfig {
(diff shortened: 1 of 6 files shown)


trosenblatt:
I don't get the link between this PR and the jira ticket described. What are those labels used for exactly ?

sds/src/rule.rs
```diff
@@ -4,6 +4,7 @@ use crate::match_action::MatchAction;
 use crate::path::Path;
 use serde_with::{serde_as, DefaultOnNull};
 
+#[serde_as]
```
vinckama, line 7: @trosenblatt 
>What are those labels used for exactly ?

`Labels`  at the rule level are like [other labels at the scanner level](https://github.com/DataDog/dd-sensitive-data-scanner/blob/main/sds/src/scanner/mod.rs#L44) to hold `key:value` for observability. They mostly refer to the term "label" defined here
https://github.com/DataDog/dd-sensitive-data-scanner/blob/main/sds/src/observability/labels.rs#L3

I decided to not expose directly `Labels`/`Label` in the RuleConfig and use instead Vec<(String, String)> so the dependency on `metrics` is fully internal to the package.

trosenblatt: But what are they used for ?

vinckama: they are used in metrics to tag metric points by org/product

trosenblatt: Ok thks I think Labels is confusing in that case why not calling it MetricLabels ?

vinckama: About renaming it `Labels` to `MetricLabels`, I'm fine doing it but at the same time I think it's fine as it is and leave a more generic name that can be re-used for trace/log labels. I can rename these in another PR if needed

trosenblatt (approved):
LGTM

PR #18, opened 2024-04-02
Sort matches before calling `calculate_indices`

fuchsnj:
https://datadoghq.atlassian.net/browse/SDS-218

Some `Encoding` implementations (e.g. `Utf16Encoding`) requires that the indices passed to it are in order, but this was not well-documented / enforced so it was broken in some recent refactoring. This is now fixed, documentation was added, and a test was added.

### DIFF
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -1035,7 +1035,7 @@ mod test {
     }
 
     #[test]
-    fn panic_test() {
+    fn test_calculate_indices_is_called_with_sorted_start_index() {
         // A custom "Event" implementation is used here to use a different encoding that asserts the indices are in order
         struct OrderAssertEvent(SimpleEvent);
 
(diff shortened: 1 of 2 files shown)


sds/src/scanner/mod.rs
```diff
@@ -1029,4 +1033,77 @@ mod test {
         // treated as a false positive.
         assert_eq!(matches.len(), 1);
     }
+
+    #[test]
+    fn panic_test() {
```
vinckama, line 1038: as this test does not panic, maybe it can be name with another name, `test_calculate_indices_is_called_with_sorted_start_index` for instance.
```suggestion
    fn test_calculate_indices_is_called_with_sorted_start_index() {
```

vinckama (approved):
LGTM with minor comment about test name

sds/src/scanner/mod.rs
fuchsnj: It did panic when I wrote it originally 😆 
I just forgot to change the name, so thanks for noticing.

PR #24, opened 2024-04-10
Fix panic

fuchsnj:
The code to retrieve the next valid char wasn't quite correct. `.char_indices()` requires starting on a valid boundary, so the `+1` could panic. Instead, the `+1` is removed, and a `nth(1)` is used to find the next valid boundary.

### DIFF
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -1181,7 +1181,7 @@ mod test {
 
         let scanner = Scanner::new(&[rule_0]).unwrap();
 
-        // The first 4 numbers match as a credit-card, but false the luhn checksum.
+        // The first 4 numbers match as a credit-card, but fail the luhn checksum.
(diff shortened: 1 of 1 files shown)


vinckama (approved):

sds/src/scanner/mod.rs
```diff
@@ -1206,4 +1206,25 @@ mod test {
...
+                replacement: "[credit card]".to_string(),
+            })
+            .validator(LuhnChecksum)
+            .build();
+
+        let scanner = Scanner::new(&[rule_0]).unwrap();
+
+        // The first 4 numbers match as a credit-card, but fail the luhn checksum.
+        // The last 4 numbers (which overlap with the first match) pass the checksum.
+        let mut content = "[5184,5185,5252,5052€5005]".to_string();
```
vinckama, line 1224: maybe let's add `€` in the regex. Also, the regex could me the same as previous one `([,€]?\\d+){4}`

PR #31, opened 2024-04-18
scanner: change the `Scan` signature and behaviour.

remeh:
## What's this PR about?

First change: `Scan` is now always returning an event value, either the unchanged event or the processed event.

Secondly, it returns a new boolean indicating if the said event has been mutated or not while going through the scanner.

The reason behind this change is that if a match were happening with a "None" action, the caller had no easy way to know if they have to use the returned slice of bytes or not, while the returned slice being empty since no **mutation** happened, despite the match.

## Remark

Another solution would've been to have the `Scan()` signature unchanged, except for using the `[]byte` to return either the unchanged or mutated event.

  1. I think it's nicer to add the information to the caller, because otherwise if the caller needs to know if the event has been mutated or not, the caller has no other way than to do a byte comparison between its input and the returned slice of bytes.
  2. On top of this, it is simplifying the signature (as in, they're identical) between `Scan`, `ScanEventsMap` and `ScanEventsList`. The other solution would have been to not returned the boolean from `Scan`, creating a discrepancy between the three.

A lasting problem is with the `ScanEventsMap` and `ScanEventsList` functions, which will return something inconsistent when `mutated == false`, but I think it was already a problem before this change.

Obviously, open to the discussion.


sds-go/go/scanner.go
```diff
@@ -115,53 +115,64 @@ func (s *Scanner) scanEncodedEvent(encodedEvent []byte) ([]byte, []RuleMatch, er
...
 
 // Scan sends the string event to the SDS shared library for processing.
 // Returned values:
-//   - the processed log if any mutation happened
+//   - whether a mutation happened or not
+//   - the resulting log (possibly be unchanged)
 //   - rules matches if any rule has matched
 //   - a possible error
-func (s *Scanner) Scan(event []byte) ([]byte, []RuleMatch, error) {
+func (s *Scanner) Scan(event []byte) (bool, []byte, []RuleMatch, error) {
```
fuchsnj, line 132: It might be worth returning a new struct here? 4 values is a lot to remember what they are / the order of them. Especially with a `bool` where the meaning might not be very clear.

fuchsnj: The `[]RuleMatch` also has enough info to determine if the event was mutated or not, but if you think a `bool` is easier that's fine.

fuchsnj (commented):
Added a couple small comments. It would be nice if this API more closely matched Rust/Java where the input value is mutated, but it's more important to keep it consistent with the Go ecosystem and I'm not very familiar with that so I don't have strong opinions here.

sds-go/go/scanner.go
remeh: Hmm you are right about the RuleMatch containing the `MatchingType`... 🤔 but it would mean a `for range` on all matches to see if it has been mutated... 
Good idea I'll return a struct indeed, way more information returned separately here 👍

fuchsnj (approved):

PR #34, opened 2024-04-27
Trosenblatt/new rule trait

trosenblatt:
Introducing RuleConfigTrait in order to be able to add Rules that are not regex based
[JIRA_LINK](https://datadoghq.atlassian.net/browse/SDS-248)

### DIFF
--- a/sds-go/rust/src/native/create_scanner.rs
+++ b/sds-go/rust/src/native/create_scanner.rs
@@ -20,13 +20,13 @@ pub extern "C" fn create_scanner(
 
         // parse the json
         let rules: Vec<Box<RuleConfig>> = serde_json::from_str(&val).unwrap();
-        let rulesWithBoxed: Vec<Box<dyn RuleConfigTrait>> = rules
+        let rules_boxed: Vec<Box<dyn RuleConfigTrait>> = rules
             .into_iter()
-            .map(|r| r as Box<dyn RuleConfig>)
+            .map(|r| r as Box<dyn RuleConfigTrait>)
             .collect();
 
         // create the scanner
-        let scanner = match Scanner::new(&rules) {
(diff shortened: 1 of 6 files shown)


sds-go/rust/src/native/create_scanner.rs
```diff
@@ -19,10 +19,14 @@ pub extern "C" fn create_scanner(
         let val = c_str.to_string_lossy();
 
         // parse the json
-        let rules: Vec<RuleConfig> = serde_json::from_str(&val).unwrap();
+        let rules: Vec<Box<RuleConfig>> = serde_json::from_str(&val).unwrap();
+        let rules_boxed: Vec<Box<dyn RuleConfigTrait>> = rules
```
trosenblatt, line 23: @fuchsnj I believe we shall move the a Vec struct everywhere for rulesConfig. WDYT ?

fuchsnj: I suspect not, but it's possible there's some fancy logic in the Rust compiler that would make it work. I'm not too worried about the conversion though.

fuchsnj: You might be able to add some generics such as `<T: Into<dyn RuleConfigTrait>>` and do the conversion in the `new` function to make it easier on the caller

trosenblatt: something like this you mean ?
    pub fn new<T: Into<dyn RuleConfigTrait>>(rules: &[T]) -> Result<Self, CreateScannerError> {
        Scanner::new_with_labels(rules, NO_LABEL)
    }
    But the Into trait requires stack based object so it does not work with dyn struct. 
    Let's chat live that would be easier most likely

PR #35, opened 2024-04-29
Change ordering of multipass v0 for backwards compatability

fuchsnj:
This is a temporary change to maintain backwards compatibility. Instead of getting all excluded matches and then finding all the exclusions, it is checked as strings are scanned, so the behavior of multipass v0 depends on the order strings are scanned.

Implementations of `Event` decide the order, so fields that tend to be excluded should normally be scanned first.

### DIFF
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -96,10 +96,15 @@ impl CompiledRuleTrait for RegexCompiledRule {
                         // Matches from excluded paths are saved and used to treat additional equal matches as false positives
                         excluded_matches.insert(content[regex_match.range()].to_string());
                     } else {
-                        match_emitter.emit(StringMatch {
-                            start: regex_match.start(),
-                            end: regex_match.end(),
-                        });
+                        // If the matched content is in `excluded_matches` it should not count as a match.
+                        // This is temporary to maintain backwards compatibility, but this should eventually happen
+                        // after all scanning is done so `excluded_matches` is fully populated.
+                        if !excluded_matches.contains(&content[regex_match.range()]) {
+                            match_emitter.emit(StringMatch {
+                                start: regex_match.start(),
+                                end: regex_match.end(),
+                            });
+                        }
                     }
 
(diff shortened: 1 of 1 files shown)


artslidd (approved):
Waiting for Vincent's review, but LGTM

sds/src/scanner/mod.rs
```diff
@@ -1036,17 +1039,21 @@ mod test {
...
         ]));
 
         let matches = scanner.scan(&mut content);
 
-        // The match from the "test" field (which is excluded) is the same as the match from "message", so it is
-        // treated as a false positive.
-        assert_eq!(matches.len(), 0);
+        // Due to the ordering of the scan (alphabetical in this case), the match from "a-match" is not
+        // excluded yet, but "z-match" is, so only 1 match is found.
+        assert_eq!(matches.len(), 1);
```
vinckama, line 1056: Minor: let's validate that the path of the match is `a-match` so it's more explicit that `z-match` is the one excluded.

sds/src/scanner/mod.rs
```diff
@@ -1036,17 +1039,21 @@ mod test {
 
         let mut content = SimpleEvent::Map(BTreeMap::from([
             (
-                "message".to_string(),
+                "a-match".to_string(),
+                SimpleEvent::String("bcdef".to_string()),
```
vinckama, line 1043: I would keep the exact same input for `a-match` and `z-match` so the only difference is the path
```suggestion
                SimpleEvent::String("abcdef".to_string()),
```

vinckama (approved):
LGTM with two minor comments

sds/src/scanner/mod.rs
fuchsnj: Fixed in a9d44ba1848f996b99c3d8a4bb83e59d8af51ba6

sds/src/scanner/mod.rs
fuchsnj: Fixed in a9d44ba1848f996b99c3d8a4bb83e59d8af51ba6

PR #39, opened 2024-05-01
Fix farmhash bug

fuchsnj:
A bug was found in the `farmhash` crate which is used for hashing in SDS.

A [fix](https://github.com/seiflotfy/rust-farmhash/pull/16) was submitted to the `farmhash` crate, but since there has been no activity on that repo for over 8 years the chance of this PR actually being merged is minimal, so I forked the repo and switched this over to the fixed version.

### DIFF
--- a/sds/src/match_action.rs
+++ b/sds/src/match_action.rs
@@ -405,4 +405,17 @@ mod test {
             })
         );
     }
+
+    #[test]
+    fn test_farmhash_bugfix() {
+        // Testing the bugfix from https://github.com/seiflotfy/rust-farmhash/pull/16
+        assert_eq!(
+            MatchAction::Hash.get_replacement(&"x".repeat(128)),
+            Some(Replacement {
+                start: 0,
+                end: 128,
+                replacement: "5170af09fd870c17".into()
+            })
+        );
+    }
 }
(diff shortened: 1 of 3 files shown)


sds/src/match_action.rs
```diff
@@ -405,4 +405,17 @@ mod test {
             })
         );
     }
+
+    #[test]
+    fn test_farmhash_bugfix() {
+        // Testing the bugfix from https://github.com/seiflotfy/rust-farmhash/pull/16
+        assert_eq!(
+            MatchAction::Hash.get_replacement(&"x".repeat(128)),
```
artslidd, line 413: IIUC, you test that for 128 (2 times 64) you end up with the correct hash value. How do you know it's the correct hash generated?

artslidd (approved):
LGTM

sds/src/match_action.rs
fuchsnj: The test value here matches the value from both the java library https://github.com/OpenHFT/Zero-Allocation-Hashing and the online tool https://md5hashing.net/hash/farmHashFingerprint64/5170af09fd870c17

I also verified the correct behavior from the original c reference implementation on this line: https://github.com/google/farmhash/blob/master/src/farmhash.cc#L558

PR #40, opened 2024-05-02
🔭 [SDS-247] Add metric regarding multi-pass v0

artslidd:


sds/src/scanner/metrics.rs
```diff
@@ -0,0 +1,25 @@
...
+                labels.clone_with_labels(Labels::new(&[(TYPE, "excluded_attributes".to_string())]))
+            ),
+        }
+    }
+}
+
+impl Default for Metrics {
+    fn default() -> Self {
+        Metrics::new(&NO_LABEL)
+    }
```
vinckama, line 24: I'm not sure it's needed, maybe we don't need a default here.

sds/src/scanner/metrics.rs
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const TYPE: &str = "type";
+
+impl Metrics {
+    pub fn new(labels: &Labels) -> Self {
+        Metrics {
+            false_positive_excluded_attributes: counter!(
+                "false_positive.excluded_attributes",
+                labels.clone_with_labels(Labels::new(&[(TYPE, "excluded_attributes".to_string())]))
```
vinckama, line 15: Minor, no need to add the `type:excluded_attributes` tag here as the same of the metric is already pretty explicit.
Also, I believe we can find a name a bit more explicit for this metric. What do you think of `false_positive.multipass.excluded_match` ?

sds/src/scanner/mod.rs
```diff
@@ -88,6 +91,7 @@ impl Scanner {
             rules: rules.clone(),
             scoped_ruleset,
             cache_pool: CachePool::new(rules),
+            metrics: Metrics::new(&scanner_labels),
```
vinckama, line 94: [`logs.sds.false_positives.v0` is tagged by `scrubbing_rule_name`](https://app.datadoghq.com/dashboard/veg-ziv-2ea/logs-sds-overview?fromUser=false&fullscreen_end_ts=1714983264788&fullscreen_paused=false&fullscreen_refresh_mode=sliding&fullscreen_section=edit&fullscreen_start_ts=1714968864788&fullscreen_widget=5222798691058174&refresh_mode=sliding&view=spans&from_ts=1714968558197&to_ts=1714982958197&live=true). With the current implementation, we will lose this tag. Would it be possible to keep this tag?

vinckama (commented):
I added a few comments

sds/src/scanner/metrics.rs
artslidd: Yes, done in [Remove Default implementation + rename metric + remove extra labels](https://github.com/DataDog/dd-sensitive-data-scanner/pull/40/commits/382e492e80b97ab6c57f6672f5e028ab21e40ad9)

sds/src/scanner/metrics.rs
artslidd: Removed this in [Remove Default implementation + rename metric + remove extra labels](https://github.com/DataDog/dd-sensitive-data-scanner/pull/40/commits/382e492e80b97ab6c57f6672f5e028ab21e40ad9)

sds/src/scanner/mod.rs
artslidd: I re-implemented this to include the tags of the rule: [Create metric for each compiled rule](https://github.com/DataDog/dd-sensitive-data-scanner/pull/40/commits/7f28c8b3ee2bc2dfa8939a69cb4cb5f0ab45cd09)

vinckama (commented):
LGTM but I think you will have some conflicts with the main branch. Let me know when they are fixed, I will take another look

sds/src/scanner/mod.rs
```diff
@@ -1237,6 +1258,27 @@ mod test {
             matches[0].path,
             Path::from(vec![PathSegment::Field("a-match".into())])
         );
+
+        let snapshot = snapshotter.snapshot().into_vec();
+        assert_eq!(snapshot.len(), 3);
+
+        let metric_name = "false_positive.multipass.excluded_match";
+        let metric_pos = snapshot
```
fuchsnj, line 1266: I would recommend converting the `snapshot` [into a HashMap](https://docs.rs/metrics-util/0.16.3/metrics_util/debugging/struct.Snapshot.html#method.into_hashmap) so you can directly lookup by key.

sds/src/scanner/mod.rs
```diff
@@ -1196,39 +1206,50 @@ mod test {
...
-                "a-match".to_string(),
-                SimpleEvent::String("bcdef".to_string()),
-            ),
-            (
-                "z-match".to_string(),
-                SimpleEvent::String("bcdef".to_string()),
-            ),
-            ("test".to_string(), SimpleEvent::String("bcdef".to_string())),
-        ]));
+        let matches = metrics::with_local_recorder(&recorder, || {
```
fuchsnj, line 1224: In general it's better to create a new test for specifically testing metrics. This unit test was already a slightly more complicated one where the order of the fields matters. It's probably better to create a new one with a more simple test to isolate the test to just the metrics testing as much as possible

artslidd: Yes definitely! typing...

artslidd: [Refactor tests](https://github.com/DataDog/dd-sensitive-data-scanner/pull/40/commits/3e65fc841e51866c89acec7e08121c0440cefca3)

sds/src/scanner/mod.rs
```diff
@@ -1434,4 +1444,145 @@ mod test {
         // This should match because "test" is not found, so it's not a false-positive
         assert_eq!(matches.len(), 1);
     }
+
+    mod metrics_test {
```
fuchsnj, line 1448: This needs `#[cfg(test)]` so it's only compiled when running tests. (This allows several things such as speeding up compilation when not running tests, and allowing using "dev-dependencies" only during tests)

artslidd: Even if the `mod` declaration is nested in a module that is already `#[cfg(test)]` ?

fuchsnj (approved):
1 small suggestion, otherwise looks good

PR #48, opened 2024-05-16
Add support for standard scanners

fuchsnj:
🚧 work in progress 🚧


sds/src/standard_rules/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+use std::fs::File;
+use std::io::BufReader;
+use std::path::Path;
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub struct StandardRule {
+    pub id: String,
+    pub name: String,
+    pub description: String,
+    pub pattern: String,
```
trosenblatt, line 14: @fuchsnj can you add a type there 

```
                if (rule.getScannerType().equals(ScannerType.CREDIT_CARD_SCANNER)) {
                  builder.secondaryValidator(SecondaryValidator.LUHN_CHECKSUM);
                } else if (rule.getScannerType().equals(ScannerType.CHINESE_ID_SCANNER)) {
                  builder.secondaryValidator(SecondaryValidator.CHINESE_ID_CHECKSUM);
                } else if (rule.getScannerType().equals(ScannerType.NAME_SCANNER)) {
                  builder.type(SdsRule.Builder.NAME_DETECTION_TYPE);
```

trosenblatt: And a bool experimental

sds/src/standard_rules/mod.rs
```diff
@@ -0,0 +1,290 @@
...
+    pub tags: Vec<String>,
+
+    /// Labels are not currently used by this library, but the data is available.
+    #[serde(default)]
+    pub labels: Vec<String>,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub enum StandardRuleValidator {
+    LuhnChecksum,
```
trosenblatt, line 36: Why not adding the other already implemented ? (ChineseId and GitHub)

sds/src/standard_rules/mod.rs
fuchsnj: I added an experimental bool field. As for the type, I'd like to add that in a separate PR. I believe all existing standard rules (unless I'm missing some) are Regex today. Adding other types will likely end up changing the structure of the rules a bit (e.g. `pattern` is no longer required). The set of changes to get the existing rules migrated is already large enough, I'd like to keep other improvement separate.

sds/src/standard_rules/mod.rs
fuchsnj: Added.

sds/src/standard_rules/mod.rs
trosenblatt: Ok for an other PR FYI I am current working on a branch that have type = name_detection ;)

trosenblatt (approved):
LGTM

PR #52, opened 2024-05-23
✨ [SDS-194] Add path_regex to ProximityKeywordsRegex

artslidd:
## Description

First PR of three that aim to match included keywords against the path of the value that matched.

This PR adds a field called `path_regex` to the `ProximityKeywordsRegex` and a function called `calculate_keyword_path_pattern` that computes the path_pattern using pattern matching.

Some tests verify that the regex generated is correct


sds/src/proximity_keywords.rs
```diff
@@ -226,11 +241,33 @@ impl Default for Metrics {
     }
 }
 
+fn compile_keywords_pattern(keyword_patterns: Vec<(Ast, Ast)>) -> (String, String) {
+    let content_pattern = Ast::Alternation(Alternation {
```
fuchsnj, line 245: I would expect the `Ast::Alternation` to be produced by the `compile_keywords_to_ast` function, since that's part of the Ast, and it doesn't look like that function is used without it anywhere else?

This function would just be calling `.to_string()`, and could probably just be inlined where it's used

artslidd: Agreed 👍

artslidd: [Inline compile_keyword_patterns fn](https://github.com/DataDog/dd-sensitive-data-scanner/pull/52/commits/94bdea5ac283933f09e3bf0f1bbcdf90141e36b4)

sds/src/proximity_keywords.rs
```diff
@@ -317,6 +346,101 @@ fn calculate_keyword_pattern(keyword: &str) -> Ast {
...
 }
 
+#[derive(Clone, Copy)]
+enum CharType {
+    Regular,
+    Uppercase,
+    Separator,
+}
+
+fn get_chars_type(prev: &char, current: &char, next: &char) -> (CharType, CharType, CharType) {
```
vinckama, line 356: `get_chars_type` is a repetition of 3 times the same function for each character. I think it's easier to read if `get_chars_type` is updated to be only applied to a single character. This method will be simplified a lot (to look like only `get_single_char_type`) and `standardize_path_chars` won't be much more complicated:

`standardize_path_chars` can look like:
```rust
...
    for chars in chars.windows(3) {
        let current: &char = &chars[1];
        let prev_char = get_chars_type(&chars[0]);
        let current_char = get_chars_type(&chars[1]);
        let next_char = get_chars_type(&chars[2]);

        match (prev_char, current_char, next_char) {
...
```
It's not pretty but I think it's easier to read

sds/src/proximity_keywords.rs
```diff
@@ -317,6 +346,101 @@ fn calculate_keyword_pattern(keyword: &str) -> Ast {
...
+where
+    F: FnMut(&char),
+{
+    let kw_length = chars.len();
+
+    if kw_length >= 1 {
+        push_character(&chars[0])
+    }
+
+    for chars in chars.windows(3) {
```
vinckama, line 387: While reading the code, it was not clear to me why this was needed at first. I think it would be easier to read if the condition was the contrary:
```suggestion
    if chars.is_empty(){
        return;
    }

    push_character(&chars[0]);
    for chars in chars.windows(3) {
```

sds/src/proximity_keywords.rs
```diff
@@ -317,6 +346,101 @@ fn calculate_keyword_pattern(keyword: &str) -> Ast {
...
+        let current = &chars[1];
+        let (prev_char, current_char, next_char) = get_chars_type(&chars[0], &chars[1], &chars[2]);
+        match (prev_char, current_char, next_char) {
```
vinckama, line 390: It seems to me that `next_char` is not used, it could probably be dropped.

I did not try myself, but maybe you can use a simple for loop on `keyword.chars()` and keep track on the previous character type (if it exist). This way, the `char_list` won't need to be created in calculate_keyword_path_pattern, and possibly `kw_length >= 2` at the end can be dropped

vinckama (commented):
I added a few general suggestions in the PR

artslidd:
@[artslidd](https://github.com/artslidd) artslidd deleted a comment from [vinckama](https://github.com/vinckama) [17 minutes ago](https://github.com/DataDog/dd-sensitive-data-scanner/pull/52#event-12945610887)

Sorry for this, I did something wrong in Intellij

Your comment was related to this one, and I fixed them two in this commit: [Improve algorithm and reduce window to 2](https://github.com/DataDog/dd-sensitive-data-scanner/pull/52/commits/44c69740dc08a3c43ce532237f5ed8f402bc78b1)

sds/src/proximity_keywords.rs
```diff
@@ -246,69 +261,83 @@ fn compile_keywords_to_ast(
...
-    if keyword
-        .chars()
-        .next()
+fn should_push_word_boundary(keyword: &str, is_end: bool) -> bool {
+    let next_char = if is_end {
+        keyword.chars().next_back()
+    } else {
+        keyword.chars().next()
+    };
+    next_char
```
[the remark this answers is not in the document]
artslidd: Yep, I agree with this

Made the change in [Simplify should_push_word_boundary](https://github.com/DataDog/dd-sensitive-data-scanner/pull/52/commits/df46a32d3e9d3180ebd8b5ab00ad93b4d7c22cd0)

sds/src/proximity_keywords.rs
artslidd: Agreed, I did the change in [Improve algorithm and reduce window to 2](https://github.com/DataDog/dd-sensitive-data-scanner/pull/52/commits/44c69740dc08a3c43ce532237f5ed8f402bc78b1)

sds/src/proximity_keywords.rs
artslidd: [Improve algorithm and reduce window to 2](https://github.com/DataDog/dd-sensitive-data-scanner/pull/52/commits/44c69740dc08a3c43ce532237f5ed8f402bc78b1)

vinckama (commented):
a few more comments

vinckama (approved):

sds/src/proximity_keywords.rs
```diff
@@ -246,71 +261,171 @@ fn compile_keywords_to_ast(
...
+{
+    let kw_length = chars.len();
+
+    if chars.is_empty() {
+        return;
+    }
+
+    push_character(&chars[0]);
+
+    for (i, chars) in chars.windows(2).enumerate() {
```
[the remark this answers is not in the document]
vinckama: It was mostly a suggestion of style to avoid having 3 variables in the match. Both works, I don't have a strong opinion here 👍

PR #53, opened 2024-05-23
✨ [SDS-194] Add validation of included keywords using the absolute path of the match

artslidd:
- Add a `sanitize` method on the Path struct, that uses the same algorithm that is done for the proximity keywords and sanitize the path to a string
- Add a false positive method that validates keywords using the path of the match on structured events.
- Add a `ScannerFeature` struct that controls whether or not we will use that new false positive method
- Add unit tests to cover the feature

### DIFF
--- a/sds/src/path.rs
+++ b/sds/src/path.rs
@@ -1,7 +1,7 @@
 use std::borrow::Cow;
 use std::fmt::{Debug, Display, Formatter};
 
-use crate::proximity_keywords::{MULTI_WORD_KEYWORDS_LINK_CHARS, UNIFIED_LINK_CHAR};
+use crate::proximity_keywords::{standardize_path_chars, UNIFIED_LINK_CHAR};
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
@@ -56,19 +56,17 @@ impl<'a> Path<'a> {
         self.segments
             .iter()
             .filter_map(|segment| match segment {
-                PathSegment::Field(field) => Some(field.to_string().to_ascii_lowercase()),
+                PathSegment::Field(field) => {
+                    let mut sanitized_segment: Vec<char> = vec![];
+                    standardize_path_chars(field.chars().collect(), |c| {
+                        sanitized_segment.push(c.to_ascii_lowercase());
+                    });
+                    return Some(sanitized_segment.iter().collect());
+                }
                 _ => None,
             })
             .collect::<Vec<String>>()
             .join(UNIFIED_LINK_CHAR.to_string().as_str())
-            .chars()
-            .map(|c| {
-                if MULTI_WORD_KEYWORDS_LINK_CHARS.contains(&c) {
-                    return UNIFIED_LINK_CHAR;
-                }
-                c
-            })
-            .collect()
     }
 }
 
@@ -144,7 +142,7 @@ mod test {
     }
 
     #[test]
-    fn test_absolute_path() {
+    fn test_sanitize_path() {
         assert_eq!(
             Path::from(vec!["hello".into(), 0.into(), "world".into()]).sanitize(),
             "hello.world"
@@ -162,6 +160,11 @@ mod test {
             ])
             .sanitize(),
             "hello.world.of.chicken"
+        );
+
+        assert_eq!(
+            Path::from(vec!["hello_world-of-".into(), "/chickens_/".into()]).sanitize(),
+            "hello.world.of-./chickens./"
         );
     }
 }
--- a/sds/src/path.rs
+++ b/sds/src/path.rs
@@ -59,7 +59,7 @@ impl<'a> Path<'a> {
                 if i != 0 {
                     sanitized_path.push(UNIFIED_LINK_CHAR);
                 }
-                standardize_path_chars(field.chars().collect(), |c| {
+                standardize_path_chars(field, |c| {
                     sanitized_path.push(c.to_ascii_lowercase());
                 });
             }
--- a/sds/src/path.rs
+++ b/sds/src/path.rs
@@ -3,6 +3,8 @@ use std::fmt::{Debug, Display, Formatter};
 
 use serde::{Deserialize, Serialize};
 
+const UNIFIED_LINK_CHAR: &str = ".";
+
 #[derive(Serialize, Deserialize, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Path<'a> {
     pub segments: Vec<PathSegment<'a>>,
@@ -51,7 +53,7 @@ impl<'a> Path<'a> {
         true
     }
 
-    pub fn to_string(&self) -> String {
+    pub fn absolute_path(&self) -> String {
         self.segments
             .iter()
             .filter_map(|segment| match segment {
@@ -59,7 +61,7 @@ impl<'a> Path<'a> {
                 _ => None,
             })
             .collect::<Vec<String>>()
-            .join(".")
+            .join(UNIFIED_LINK_CHAR)
     }
 }
 
--- a/sds/src/path.rs
+++ b/sds/src/path.rs
@@ -52,7 +52,7 @@ impl<'a> Path<'a> {
         true
     }
 
-    pub fn absolute_path(&self) -> String {
+    pub fn sanitize(&self) -> String {
         self.segments
             .iter()
             .filter_map(|segment| match segment {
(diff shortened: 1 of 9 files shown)


sds/src/proximity_keywords.rs
```diff
@@ -127,6 +135,12 @@ impl CompiledProximityKeywords {
     }
 }
 
+fn contains_keyword_in_path(path: &str, regex: &ProximityKeywordsRegex<false>) -> bool {
```
vinckama, line 138: Another way to look at this is to use a more oriented object code. In that sense, you can implement `ProximityKeywordsRegex` with two methods: `contains_keywords_in_content` and `contains_keywords_in_path`. 

```rust
impl<const EXCLUDED_CHARS: bool> ProximityKeywordsRegex<EXCLUDED_CHARS> {
    fn contains_keywords_in_content(
        &self,
        content: &str,
        match_start: usize,
        look_ahead_char_count: usize,
    ) -> bool {
      ...
   }

   fn contains_keyword_in_path(&self, path: &str) -> bool {
       ...
   }
```

All the logic related to the match will be encapsulated in the `ProximityKeywordsRegex`. It will be less likely to use `contains_keywords_in_content` with the regex associated to the `path` and vice versa.

sds/src/proximity_keywords.rs
```diff
@@ -356,7 +370,7 @@ fn get_char_type(c: &char) -> CharType {
     }
 }
 
-fn standardize_path_chars<F>(chars: Vec<char>, mut push_character: F)
+pub fn standardize_path_chars<F>(chars: Vec<char>, mut push_character: F)
```
vinckama, line 373: As this method becomes public, it's a good practice to write some rust doc (starting with `///`) to describe what the goal of the method is.

sds/src/scanner/mod.rs
```diff
@@ -832,6 +855,95 @@ mod test {
...
+                replacement: "[REDACTED]".to_string(),
+            })
+            .proximity_keywords(ProximityKeywordsConfig {
+                look_ahead_character_count: 30,
+                included_keywords: vec!["awsAccess".to_string(), "access/key".to_string()],
+                excluded_keywords: vec![],
+            })
+            .build();
+
+        let scanner = Scanner::new(&[redact_test_rule]).unwrap();
```
vinckama, line 872: As the `should_keywords_match_event_paths` parameter is pretty important here, I think the test should make it explicit that `should_keywords_match_event_paths` is expected to be true in this test. I think it would be interesting to have also a test where `should_keywords_match_event_paths` is false.

sds/src/scanner/mod.rs
```diff
@@ -832,6 +855,95 @@ mod test {
         assert_eq!(matches.len(), 1);
     }
 
+    #[test]
+    fn test_included_keywords_match_path() {
```
vinckama, line 860: As a general guideline,  I think it's better to either have more unitary test.
Either each case in this test should be explicitly described with a comment or even better each case has is dedicated test with a very explicit name. It's way easier to debug/update/fix when they fail following a change of behaviour. It will also better scope the test and avoid having duplicated tests

Here, I would recommend splitting this test in 5 tests:
- test_included_keywords_should_match_when_find_in_path
- test_included_keywords_should_match_when_find_in_path_with_case_insensivity
- test_included_keywords_should_not_match_when_missing_in_path
...

sds/src/scanner/mod.rs
```diff
@@ -832,6 +856,95 @@ mod test {
         assert_eq!(matches.len(), 1);
     }
 
+    #[test]
```
vinckama, line 859: It's possible that the path is missing in case a Event is a String. you can probably add a test for this case.

sds/src/proximity_keywords.rs
```diff
@@ -32,29 +32,37 @@ const EXCLUDED_KEYWORDS_REMOVED_CHARS: &[char] = &['-', '_'];
...
+                }
+
+                let is_valid_from_content = contains_keyword_match(
                     content,
                     match_start,
                     self.look_ahead_character_count,
                     included_keywords,
                 );
+
+                let is_false_positive = !(is_valid_from_path || is_valid_from_content);
```
vinckama, line 64: Here `is_valid_from_path` will be alway false, it can probably be simplified.

sds/src/scanner/mod.rs
```diff
@@ -48,6 +48,7 @@ pub trait CompiledRuleTrait: Send + Sync {
     fn get_string_matches(
         &self,
         content: &str,
+        path: Option<&str>,
```
vinckama, line 51: As this is not the path, I think it should be named differently.
Also, there is possibly no need to need to sanitize the path on each `get_string_matches`, this will be needed only if we go in `is_false_positive_match`. Another option here could be to expose the real path in `get_string_matches` and to sanitise it only in `contains_keyword_in_path` when it's needed.

Here the advantage is that you will need to sanitise the path exactly once for each field (and not multiple time if there is multiple match). To avoid recomputing the path multiple times for a same field, you can possibly use a singleton pattern.

vinckama (commented):
I added a few comments/questions/suggestions

sds/src/scanner/mod.rs
artslidd: Yes I agree 👍

sds/src/scanner/mod.rs
```diff
@@ -225,6 +233,9 @@ impl Scanner {
             rules: compiled_rules,
             scoped_ruleset,
             cache_pool: cache_pool_builder.build(),
+            feature_set: ScannerFeatures {
+                should_keywords_match_event_paths,
```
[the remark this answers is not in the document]
artslidd: Yep, sounds good to me 

[Expose the feature set directly in the arguments](https://github.com/DataDog/dd-sensitive-data-scanner/pull/53/commits/14bdfcc1531898602ce6336de50f6b7aaea66aed)

sds/src/proximity_keywords.rs
artslidd: [Remove useless variable](https://github.com/DataDog/dd-sensitive-data-scanner/pull/53/commits/59b7c183a9654134a6cd009f088310c5671745fb)

sds/src/scanner/mod.rs
artslidd: I added a test case corresponding to when the feature is disabled: [Split tests and move path sanitization](https://github.com/DataDog/dd-sensitive-data-scanner/pull/53/commits/4c3068390ed5fce0bcadc0f6471fa1586bda910e)

sds/src/scanner/mod.rs
artslidd: Yep, that's a good idea
[Split tests and move path sanitization](https://github.com/DataDog/dd-sensitive-data-scanner/pull/53/commits/4c3068390ed5fce0bcadc0f6471fa1586bda910e)

sds/src/scanner/mod.rs
artslidd: I did it in [Split tests and move path sanitization](https://github.com/DataDog/dd-sensitive-data-scanner/pull/53/commits/4c3068390ed5fce0bcadc0f6471fa1586bda910e)

sds/src/proximity_keywords.rs
artslidd: [Add documentation for standardize_path_chars](https://github.com/DataDog/dd-sensitive-data-scanner/pull/53/commits/8f03a2114bf2bd03e8042f6cd2354cd1ecd16792)

sds/src/proximity_keywords.rs
artslidd: [Move functions to methods to have more resilient code](https://github.com/DataDog/dd-sensitive-data-scanner/pull/53/commits/f30896ea6cefd294e7bc98bfb9d53810db397a81)

sds/src/path.rs
```diff
@@ -50,6 +51,23 @@ impl<'a> Path<'a> {
         }
         true
     }
+
+    pub fn sanitize(&self) -> String {
```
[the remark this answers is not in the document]
artslidd: Do you think this is a good refactoring? [Improve sanitize algorithm](https://github.com/DataDog/dd-sensitive-data-scanner/pull/53/commits/18216516b958e2fd793230fb709bd380189cde16)

sds/src/scanner/mod.rs
```diff
@@ -186,20 +195,29 @@ impl RuleConfigTrait for RegexRuleConfig {
     }
 }
 
+pub struct ScannerFeatures {
```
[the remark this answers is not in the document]
artslidd: I'm tackling this in another PR as this one is already big

sds/src/path.rs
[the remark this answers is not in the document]
fuchsnj: The changes you made are all good. The `field.chars().collect()` is still going to generate quite a few unnecessary memory allocations though. I would refactor `standardize_path_chars` to accept a `&str` instead of `Vec<char>`, and that should improve things quite a bit there.

The `standardize_path_chars` implementation was fine when it was just used for generating the regex during scanner creation, but now it's used during scanning which is much more performance sensitive.

sds/src/scanner/mod.rs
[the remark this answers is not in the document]
artslidd: Done in https://github.com/DataDog/dd-sensitive-data-scanner/pull/59/files

sds/src/path.rs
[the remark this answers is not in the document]
artslidd: Ok I see, I'm doing the change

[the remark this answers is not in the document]
artslidd: [Avoid using Vec[char] in path standardization](https://github.com/DataDog/dd-sensitive-data-scanner/pull/53/commits/53900ad8afc3cefd8f0b8bc1cba00333a0b1292d)

I'm merging to be able to test on staging today, but feel free to tell me if something can still be improved here 🙇

PR #66, opened 2024-06-10
Fail CI on (Rust) warnings

fuchsnj:
[Jira issue](https://datadoghq.atlassian.net/jira/software/projects/SDS/boards/5414?selectedIssue=SDS-313)

CI now fails on Rust warnings.

Test failure: https://github.com/DataDog/dd-sensitive-data-scanner/actions/runs/9449369037/job/26025470990?pr=66#step:6:262

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,7 @@ name: Test Suite
 on:
   pull_request:
   merge_group:
-    types: [checks_requested]
+    types: [ checks_requested ]
   push:
     branches:
       - main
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -21,6 +21,7 @@ env:
   VERBOSE: true
   CI: true
   PROFILE: debug
+  RUSTFLAGS: "-D warnings"
```
vinckama, line 24: I think it would be great to link the [doc](https://doc.rust-lang.org/rustc/lints/groups.html) in the PR description, so it is easier for a next reader to understand what is flag mean.

vinckama (approved):
LGTM with a minor comment on the PR description

.github/workflows/test.yml
fuchsnj: Fixed

PR #67, opened 2024-06-10
Add redactionType to sds-go bindings after scanned event

trosenblatt:
[PR](https://datadoghq.atlassian.net/browse/SDS-301)
Parse redactionType coming from the rust scan and exposes it in sds-go binding

### DIFF
--- a/sds-go/go/scanner_test.go
+++ b/sds-go/go/scanner_test.go
@@ -265,6 +265,7 @@ func TestScanStringEvent(t *testing.T) {
 			str:     "this is a hello event",
 			rules: []RuleMatch{{
 				RuleIdx:           0,
+				ReplacementType:   ReplacementTypeNone,
 				StartIndex:        10,
 				EndIndexExclusive: 15,
 				ShiftOffset:       0,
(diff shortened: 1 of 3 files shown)


sds-go/go/scanner_test.go
```diff
@@ -462,6 +482,43 @@ func TestPartialRedact(t *testing.T) {
...
+				StartIndex:        10,
+				EndIndexExclusive: 16,
+				ShiftOffset:       0,
+			}},
+		},
+	}
+
+	runTest(t, scanner, testData)
+}
+
```
remeh, line 495: Looks like we're missing a test for Hash redaction (and thus to test `ReplacementTypeHash`), could I ask you to add one, please?

remeh (approved):
LGTM! Just mentioned a nit.

sds-go/go/scanner_test.go
trosenblatt: yes good catch

trosenblatt: the only problem is to not test the hash :)

trosenblatt: done

PR #68, opened 2024-06-17
Improve performance for rules with included keywords

fuchsnj:
Previously rules with included keywords would search for matches of the normal rule regex first, and then check if it's a false positive by looking for included keywords after a match. There are some edge cases where this can cause really bad performance, such as when the regex has a lot of matches, but the included keywords don't exist.

Now this search is reversed, only when included keywords are present. Included keywords are found first, and then a regex search is done after that. Benchmarks show that this improves perf by up to ~100x in some cases.

Because of this change, we are no longer able to emit the "false positive" metric for included keywords, so that was removed.


sds/src/proximity_keywords/excluded_keywords.rs
```diff
@@ -0,0 +1,22 @@
+use metrics::Counter;
+
+pub struct CompiledExcludedProximityKeywords {
+    pub look_ahead_character_count: usize,
+    pub keywords_pattern: super::ProximityKeywordsRegex,
```
vinckama, line 5: I don't really understand why the `super::` part is needed here. What is the advantage of doing that compared to importing the struct? 🙏

sds/src/proximity_keywords/mod.rs
```diff
@@ -12,191 +18,128 @@ use regex_syntax::ast::{
...
-            let prefix_start_info = get_prefix_start(
-                match_start,
-                // Adding 1 to the start to account for assertion checking
-                look_ahead_char_count + 1,
-                content,
-            );
+/// Returns the match context which is what is searched for keywords
+/// and the range where matches are searched for. The range is needed since the context is
+/// expanded to ensure regex assertions (e.g. word boundaries) work correctly.
+fn contains_keyword_match(
```
vinckama, line 109: This method is now only used in excluded keywords. It's also specific to them due to the `EXCLUDED_KEYWORDS_REMOVED_CHARS` replacement line 127. Would it make sense to move it in `excluded_keywords.rs`?

vinckama (commented):
I reviewed most of the PR, I still need to check the tests. I added two comments so far.

sds/src/proximity_keywords/excluded_keywords.rs
fuchsnj: It's not needed. Probably just a side-effect of the IDE importing things for me, and lots of refactoring while writing this. I'll remove it.

sds/src/proximity_keywords/mod.rs
fuchsnj: https://github.com/DataDog/dd-sensitive-data-scanner/pull/68/commits/8be9f95221846052b7352169b7d53a24e9375e90

fuchsnj:
> In my opinion, the logic is good but the structure is a bit hard to follow

The first implementation I had of this was much worse too. I did at least start to help this by implementing 2 iterators (for included keyword search, and true positive search) which helped clean up the logic a lot, and I started splitting more of the logic out into different files (keywords, regex match). I'll continue to try to clean things up as I touch stuff.

sds/src/proximity_keywords/included_keywords.rs
```diff
@@ -0,0 +1,261 @@
...
+
+    #[test]
+    pub fn test_included_keyword_search() {
+        let keywords = super::CompiledIncludedProximityKeywords {
+            look_ahead_character_count: 0,
+            keywords_pattern: ProximityKeywordsRegex {
+                content_regex: meta::Regex::new(r"\b(foo|bar)\b").unwrap(),
+                path_regex: meta::Regex::new(r"unused").unwrap(),
+            },
+        };
```
vinckama, line 84: I think it can be simplified a bit:
```suggestion
        let keywords = compile_keywords(4, &["foo", "bar"]);
```


but it also seems that this test overlap with `should_detect_on_any_keyword` so possibly we can get rid of it

sds/src/scanner/mod.rs
```diff
@@ -1718,6 +1648,175 @@ mod test {
...
+            .with_keywords_should_match_event_paths(true)
+            .build()
+            .unwrap();
+
+        let mut content = "hello [this block is exactly 37 chars long] world".to_string();
+        let matches = scanner.scan(&mut content);
+
+        // There should be no matches because the included keyword is too far from the actual match
+        assert_eq!(matches.len(), 0);
+    }
```
vinckama, line 1674: I think it's a really good example of descriptive test:
- the name is clear
- `"hello [this block is exactly 37 chars long] world"` is super explicit 💯 

a small nipick:
you can possibly drop the comment by changing the test name:
test_included_keyword_**not**_match_further_than_look_ahead_character_count

sds/src/scanner/mod.rs
```diff
@@ -1718,6 +1648,175 @@ mod test {
...
+            })
+            .build()])
+        .build()
+        .unwrap();
+
+        // only the included keyword is present
+        let mut content = "hey world".to_string();
+        assert_eq!(scanner.scan(&mut content).len(), 1);
+
+        // // only the excluded keyword is present
```
vinckama, line 1807:
```suggestion
        // only the excluded keyword is present
```

vinckama (approved):
A few minor comments

PR #69, opened 2024-06-24
Fix farmhash override

fuchsnj:
`[patch]` sections are ignored when used as a dependency, so the fix was only used locally for tests, and didn't actually fix the issue.

Previous hash mismatches that were fixed were likely due to different underlying data that was hashed (other mismatch fixes), and not the hash function itself.


trosenblatt (approved):
LGTM

PR #71, opened 2024-06-25
Trosenblatt/rule match cache

trosenblatt:
This adds a rule_idx and rule_match_cache to the rule get_string_match so that name_detection rule will be able to use it to only parse "message" field and reuse matches to detect name in sub attributes

[JIRA_LINK](https://datadoghq.atlassian.net/browse/SDS-335)


sds/src/scanner/mod.rs
```diff
@@ -53,6 +54,8 @@ pub trait CompiledRuleTrait: Send + Sync {
         content: &str,
         path: &Path,
         caches: &mut CachePoolGuard<'_>,
+        rule_idx: usize,
+        cached_string_matches_per_rule_idx: &mut HashMap<usize, AHashSet<String>>,
```
vinckama, line 58: If only the entry of the cache associated to the `rule_idx` will be used, would it make sense to only keep this entry in the `get_string_matches` method, in other words:
```suggestion
        cached_string_matches: &mut AHashSet<String>,
```

sds/src/simple_event.rs
```diff
@@ -2,7 +2,10 @@ use std::collections::BTreeMap;
 
 use crate::{encoding::Utf8Encoding, Event, EventVisitor, Path, PathSegment};
 
+static FIRST_VISIT_KEY: &str = "message";
+
 /// A simple implementation of `Event`. This is meant for testing / demonstration purposes.
+/// It always start visiting elements with key == "message" first, then the rest of the keys in the map
```
vinckama, line 8: multi-pass v0 rely on ordering as well for now and expect the path to be scanned last. This will also break the implicit expectation related to filtered encoded fields [here](https://github.com/DataDog/sds-shared-library/blob/main/sds-java/src/main/java/dd/sds/Encoder.java#L52-L69).

vinckama (commented):
I added two comments

sds/src/simple_event.rs
vinckama: This constraint might disappear in the future as cleanup of the java migration (multi-pass won't rely anymore on ordering) but it's a hard blocker for now

vinckama: More generally, we should not have any assumption in the sds-engine on the input structure. In fact, the `message` field  in logs can be named differently in spans/rum events. In my opinion, this constraint, if we really want it, should  to defined a scanner option configured by users, not as a hard-coded argument.

vinckama: resolving this comment as I realised that this is only used for tests and not anywhere else

vinckama (approved):

PR #74, opened 2024-07-05
✨ Improve performance of `sanitize_path` and `standardize_path_chars`

artslidd:
Improve perf with two small improvements:

- [Add with_capacity to sanitize_path](https://github.com/DataDog/dd-sensitive-data-scanner/commit/02279cd1328b1867fb6a2d770ef45a9388fb0983) (we know an upper bound of the size of the string in advance)
- [Improve standardize_path_chars by early pushing characters if segment…](https://github.com/DataDog/dd-sensitive-data-scanner/commit/0ea0d9c40d4c56cbbdbffd0f4fb51c14d4abd77e). Push characters early in the method if all are lowercase or all are uppercase

### DIFF
--- a/sds/src/path.rs
+++ b/sds/src/path.rs
@@ -1,7 +1,9 @@
 use std::borrow::Cow;
 use std::fmt::{Debug, Display, Formatter};
 
-use crate::proximity_keywords::{should_bypass_standardize_path, standardize_path_chars, UNIFIED_LINK_CHAR};
+use crate::proximity_keywords::{
+    should_bypass_standardize_path, standardize_path_chars, UNIFIED_LINK_CHAR,
+};
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
@@ -179,7 +181,7 @@ mod test {
                 "CHICKEN".into(),
                 2.into(),
             ])
-                .sanitize(),
+            .sanitize(),
             "hello.world.of.chicken"
         );
 
--- a/sds/src/path.rs
+++ b/sds/src/path.rs
@@ -52,7 +52,7 @@ impl<'a> Path<'a> {
         true
     }
 
-    fn size(&self) -> usize {
+    fn size_segments_only(&self) -> usize {
         self.segments
             .iter()
             .map(|segment| {
@@ -65,7 +65,8 @@ impl<'a> Path<'a> {
     }
 
     pub fn sanitize(&self) -> String {
-        let mut sanitized_path = String::with_capacity(self.size() * 2);
+        let size_segments = self.size_segments_only();
+        let mut sanitized_path = String::with_capacity(size_segments + size_segments / 2);
         self.segments.iter().enumerate().for_each(|(i, segment)| {
             if let PathSegment::Field(field) = segment {
                 if i != 0 {
@@ -173,7 +174,7 @@ mod test {
                 "CHICKEN".into(),
                 2.into(),
             ])
-            .sanitize(),
+                .sanitize(),
             "hello.world.of.chicken"
         );
 
@@ -186,11 +187,11 @@ mod test {
     #[test]
     fn test_size() {
         assert_eq!(
-            Path::from(vec!["hello".into(), 0.into(), "world".into()]).size(),
+            Path::from(vec!["hello".into(), 0.into(), "world".into()]).size_segments_only(),
             10
         );
         assert_eq!(
-            Path::from(vec!["".into(), 0.into(), "path✅".into()]).size(),
+            Path::from(vec!["".into(), 0.into(), "path✅".into()]).size_segments_only(),
             7
         );
     }
--- a/sds/src/path.rs
+++ b/sds/src/path.rs
@@ -1,7 +1,7 @@
 use std::borrow::Cow;
 use std::fmt::{Debug, Display, Formatter};
 
-use crate::proximity_keywords::{standardize_path_chars, UNIFIED_LINK_CHAR};
+use crate::proximity_keywords::{should_bypass_standardize_path, standardize_path_chars, UNIFIED_LINK_CHAR};
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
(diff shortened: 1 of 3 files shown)


sds/src/proximity_keywords/mod.rs
```diff
@@ -279,19 +279,29 @@ fn get_char_type(c: &char) -> CharType {
...
 
-    if let Some(c) = chars.chars().next() {
-        push_character(&c);
-    } else {
+    let mut char_iterator = characters.chars();
+    if char_iterator.all(|c| c.is_ascii_lowercase())
+        || char_iterator.all(|c| c.is_ascii_uppercase())
+    {
+        for c in characters.chars() {
+            push_character(&c)
```
vinckama, line 296: The method push_character can possibly be updated to push string slice. This should then be a bit faster

vinckama (commented):
I added two suggestions. As you discuss it with @fuchsnj, I leave the "final" review to him

sds/src/proximity_keywords/mod.rs
```diff
@@ -279,19 +279,29 @@ fn get_char_type(c: &char) -> CharType {
...
+        }
         return;
     }
 
-    for (i, chars) in chars.as_bytes().windows(2).enumerate() {
+    let character_bytes = characters.as_bytes();
```
fuchsnj, line 301: Why is `characters` being converted to bytes, and then back to `chars`? (It was already like this before but I didn't notice it). This seems both unnecessary and potentially dangerous since it's assuming all characters are a single byte.

sds/src/proximity_keywords/mod.rs
fuchsnj: If I'm reading it right, this is just a short-cut to prevent doing the full sanitization in some common cases? In that case this should happen before the String allocation in `sanitize()`. At a minimum this can just call `to_owned()` on the input string (so it allocates the exact amount needed, and this be more likely to use an efficient mem-copy implementation). Depending on how this sanitized path is used, you could consider returning a `Cow<str>` which will allow returning either the original input slice (no copy needed) or a `String` (if it needs to be modified).

sds/src/path.rs
```diff
@@ -52,8 +52,20 @@ impl<'a> Path<'a> {
         true
     }
 
+    fn size(&self) -> usize {
```
fuchsnj, line 55: If this is skipping non-Field segments, the function name should probably be more specific than `size()`. I would probably assume by just the function name that this is the number of segments in a path, not the sum of all of the bytes of fields.

sds/src/path.rs
```diff
@@ -52,8 +52,20 @@ impl<'a> Path<'a> {
...
+                    return field.len();
+                }
+                0
+            })
+            .sum()
+    }
+
     pub fn sanitize(&self) -> String {
-        let mut sanitized_path = "".to_string();
+        let mut sanitized_path = String::with_capacity(self.size() * 2);
```
fuchsnj, line 68: `*2` seems a bit much. How did the benchmarks look when you changed just this 1 line? We could probably get away with a smaller number, maybe 1.5 or so. This doesn't necessarily have to be the upper bound, if the capacity is set too small it will still grow automatically as needed. It would be interesting to know how much time is spent in this allocation vs the rest of the sanitization code to be able to fine-tune this better. It could end up being worth doing a quick scan to get the exact size need (especially if that scan is combined with the check to skip sanitization all-together)

sds/src/path.rs
artslidd: I've renamed the function to be `size_only_segments` 

[Rename size() in size_only_segments()](https://github.com/DataDog/dd-sensitive-data-scanner/pull/74/commits/b0e3c9ebcb37c2598615d5a346d2112e3b0c3ecf)

sds/src/path.rs
artslidd: I've used 1.5 times, thanks for the comment :)

sds/src/proximity_keywords/mod.rs
```diff
@@ -279,19 +279,29 @@ fn get_char_type(c: &char) -> CharType {
...
+    if kw_length == 0 {
+        return;
+    }
 
-    if let Some(c) = chars.chars().next() {
-        push_character(&c);
-    } else {
+    let mut char_iterator = characters.chars();
+    if char_iterator.all(|c| c.is_ascii_lowercase())
+        || char_iterator.all(|c| c.is_ascii_uppercase())
```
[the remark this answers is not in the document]
artslidd: I've improved this in [Add should_bypass_standardize fn](https://github.com/DataDog/dd-sensitive-data-scanner/pull/74/commits/4171144bca5889d43d76b98d693b4f85ec310b6e)

sds/src/proximity_keywords/mod.rs
artslidd: I've done this: 
[Add should_bypass_standardize fn](https://github.com/DataDog/dd-sensitive-data-scanner/pull/74/commits/4171144bca5889d43d76b98d693b4f85ec310b6e) and
[Use should_bypass in standardize_path](https://github.com/DataDog/dd-sensitive-data-scanner/pull/74/commits/5fb83ecdc9e031678da9a0d72f9f18f889a2ddc4)

Tell me what you think :)

sds/src/proximity_keywords/mod.rs
artslidd: Yep! I agree it's dangerous: [Remove .as_bytes and use the iterator instead](https://github.com/DataDog/dd-sensitive-data-scanner/pull/74/commits/329ec79641dd5e707c6ef3347071fd19f2d065fc)

artslidd:
@fuchsnj I also added a benchmark: [Add benchmark](https://github.com/DataDog/dd-sensitive-data-scanner/pull/74/commits/66ab398fe9215d5240afb21e3d3348ae24d96666)

fuchsnj:
> @fuchsnj I also added a benchmark: [Add benchmark](https://github.com/DataDog/dd-sensitive-data-scanner/pull/74/commits/66ab398fe9215d5240afb21e3d3348ae24d96666)

Awesome. Do you have numbers from that from before/after this change?

artslidd:
> > @fuchsnj I also added a benchmark: [Add benchmark](https://github.com/DataDog/dd-sensitive-data-scanner/pull/74/commits/66ab398fe9215d5240afb21e3d3348ae24d96666)
> 
> Awesome. Do you have numbers from that from before/after this change?

They're not very impressive (it doesn't say that it improves the performance for sure)

What I saw though is that it takes ~5x more time with the feature enabled than with the feature disabled

fuchsnj (approved):

PR #75, opened 2024-07-05
Adding iban checksum

trosenblatt:
[JIRA_LINK](https://datadoghq.atlassian.net/browse/SDS-171)

Relying on https://docs.rs/iban_validate/4.0.1/iban/struct.Iban.html

It support both [checksum validation](https://github.com/ThomasdenH/iban_validate/blob/master/iban_validate/src/base_iban.rs)  and bban [checksum](https://github.com/ThomasdenH/iban_validate/blob/master/iban_validate/src/lib.rs)

### DIFF
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -551,6 +551,7 @@ mod test {
     use crate::validation::RegexValidationError;
     use crate::SecondaryValidator::ChineseIdChecksum;
     use crate::SecondaryValidator::GithubTokenChecksum;
+    use crate::SecondaryValidator::IbanChecker;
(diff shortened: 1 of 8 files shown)


sds/src/scanner/mod.rs
```diff
@@ -1094,6 +1095,52 @@ mod test {
         assert_eq!(content, "[IDCARD] 513231200012121651");
     }
 
+    #[test]
+    fn test_iban_checksum() {
```
vinckama, line 1099: As for NHS checksum, this can be simplified a bit to validate that the iban/bban are validated when `IbanChecker` is supplied.

vinckama (approved):
LGTM with minor comment

PR #78, opened 2024-07-15
Create GroupCache and GroupCacheConfig so that rules can have an access a shared cache by type

trosenblatt:


sds/src/scanner/cache_pool.rs
```diff
@@ -31,10 +34,31 @@ impl CachePool {
     }
 }
 
+impl GroupCacheTrait for CachePool {
```
fuchsnj, line 37: If this trait doesn't have anything other than the `any` methods, you can probably just directly use the `Any` trait instead. (e.g. return `Box<dyn Any>` instead of `Box<dyn GroupCacheTrait>`)

trosenblatt: Ok good to know. That being said it will probably have a prepare_scan_content or similar
- for ML Rules to compute features
- for Regex to get the PoolGuard (once I understand how to retain a lifetime PoolGuard in a data structure with a different lifetime)

PR #79, opened 2024-07-15
Make multi-pass V0 order independent (again)

fuchsnj:
[Jira](https://datadoghq.atlassian.net/browse/SDS-386)

This re-enables the order-independent behavior of Multi-pass V0. All excluded matches are collected before matches are checked if they should be excluded.

Between the time this was disabled and now, support for multiple rule types was added, but multipass V0 was kept only for regex based rules. To keep the behavior the same here and _only_ make it order-independent, 2 new functions were added to the rule trait, so rules can opt-in to multi-pass V0. This is enabled for regex rules, and currently defaults to false for all others.

### History
- https://github.com/DataDog/dd-sensitive-data-scanner/pull/10
- https://github.com/DataDog/dd-sensitive-data-scanner/pull/35
- This PR


trosenblatt (approved):
LGTM. 
Thanks for having done this quickly
Don't forget the Todo to report increase metrics

PR #83, opened 2024-07-22
Add GroupCache per content scanned so that ML Feature computation can ben stored

trosenblatt:
[JIRA_LINK](https://datadoghq.atlassian.net/browse/SDS-385)


sds/src/scanner/mod.rs
```diff
@@ -44,30 +45,107 @@ where
...
+        excluded_matches: &mut AHashSet<String>,
+        match_emitter: &mut dyn MatchEmitter,
+        should_keywords_match_event_paths: bool,
+    );
+}
+
+pub trait CompiledRule: Send + Sync {
+    /// Data that is instantiated once per string being scanned, and shared with all rules that
+    /// have the same `GroupData` type. `Default` will be used to initialize this data.
+    type GroupData: Default + 'static;
```
vinckama, line 69: I'm not clear on the motivation of this change and on reasons why it is needed. Would it be possible to add this `GroupData` only in the ML rule implementing `CompiledRuleDyn`?

sds/src/scanner/mod.rs
```diff
@@ -101,9 +179,9 @@ impl Scanner {
             ScannerContentVisitor {
                 scanner: self,
                 caches,
+                // cache_per_type: &mut cache_per_type,
```
vinckama, line 182: This can probably be removed
```suggestion
```

sds/src/scanner/mod.rs
```diff
@@ -487,10 +572,15 @@ mod test {
...
-    impl CompiledRuleTrait for DumbCompiledRule {
+    impl CompiledRule for DumbCompiledRule {
+        type GroupData = ();
+
         fn get_match_action(&self) -> &MatchAction {
             &self.match_action
         }
+        // fn get_cache_type(&self) -> GroupCacheType {
+        //     GroupCacheType::None
+        // }
```
vinckama, line 583: Same on this comment
```suggestion
```

sds/src/scanner/mod.rs
```diff
@@ -44,30 +45,107 @@ where
     }
 }
 
-pub trait CompiledRuleTrait: Send + Sync {
+pub trait CompiledRuleDyn: Send + Sync {
```
vinckama, line 48: Would the previous name still make sense with the current update? (or maybe why `Dyn` keyword better fit now compared to `Trait`)

vinckama (commented):
I added a few comments. I have a general question on the motivation of the PR. I added it in a comment so it can be discussed in a thread.

sds/src/scanner/mod.rs
trosenblatt: This is a rust trick that @fuchsnj did to be able to allocate a GroupData only for the CompiledRule that has a GroupData type associated with it.
For Regex
impl CompiledRule for RegexCompiledRule {
    // no special data
    type GroupData = ();
    ...
    }
    
    
    So we don't add entry to hashmap for regex and it does not introduce any performance penalty
    
    The CompiledRuleDyn is a trick to downcast the the right groupData to its specific structure and call the CompiledRule specific implementation

vinckama: Ok it makes more sense now. Would it be possible to add that in the rustdoc of `CompiledRuleDyn` and `CompiledRule`? I think it will be easier to understand the motivation of this structure for future readers. (Usually I think code is self-explained but this time, I believe some comments would really help [because the code does some "convolutions" to find its way])

vinckama (approved):
LGTM with some comments added to explain why `CompiledRuleDyn`/`CompiledRule` are needed (non-blocking)

sds/src/scanner/mod.rs
trosenblatt: sure let me do that

trosenblatt:
Added comments. @vinckama I will merge now

PR #88, opened 2024-07-30
Add NIR checksum

vinckama:
Related to this [ticket](https://datadoghq.atlassian.net/browse/SDS-186)

### DIFF
--- a/sds/src/secondary_validation/nir_checksum.rs
+++ b/sds/src/secondary_validation/nir_checksum.rs
@@ -8,36 +8,39 @@ impl Validator for NirChecksum {
     fn is_valid_match(&self, regex_match: &str) -> bool {
         let mut input_iter = regex_match.chars();
 
-        let digit: Option<i64> = get_next_digit_chars(&mut input_iter, 13)
-            .and_then(|chars| chars.into_iter().collect::<String>().parse().ok());
+        let digit = get_next_digit_chars(&mut input_iter, 13);
         if digit.is_none() {
             return false;
         }
 
-        let checksum: Option<u8> = get_next_digit_chars(&mut input_iter, 2)
-            .and_then(|chars| chars.into_iter().collect::<String>().parse().ok());
-        if checksum.is_none() || get_next_digit_str(&mut input_iter).is_some() {
+        let checksum = get_next_digit_chars(&mut input_iter, 2);
+        if checksum.is_none() || get_next_digit(&mut input_iter).is_some() {
             return false;
         }
 
-        (97 - digit.unwrap() % 97) as u8 == checksum.unwrap()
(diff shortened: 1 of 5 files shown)


sds/src/secondary_validation/nir_checksum.rs
```diff
@@ -0,0 +1,75 @@
...
+            return false;
+        }
+
+        let checksum: Option<u8> = get_next_digit_chars(&mut input_iter, 2)
+            .and_then(|chars| chars.into_iter().collect::<String>().parse().ok());
+        if checksum.is_none() || get_next_digit_str(&mut input_iter).is_some() {
+            return false;
+        }
+
+        (97 - digit.unwrap() % 97) as u8 == checksum.unwrap()
```
fuchsnj, line 23: Where does `97` come from? If it's the ASCII value for `a`, it should probably be a constant like `const A_VALUE: u8 = 'a' as u8;`

sds/src/secondary_validation/nir_checksum.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+fn get_next_digit_str(chars: &mut Chars<'_>) -> Option<char> {
+    chars.find(|&char| char.is_ascii_digit())
+}
+
+fn get_next_digit_chars(chars: &mut Chars<'_>, size: usize) -> Option<Vec<char>> {
```
fuchsnj, line 31: This function seems to be unnecessarily allocating. Chars are collected and put in a `Vec`, which is immediately iterated over again and collected into a `String`, which is then immediately used to converted into a number. Ideally this would directly compute the integer representation do no allocation is needed at all.

If that is too much to do right now, the chars could at least be collected immediately into a `String` and that is returned instead of the intermediate `Vec`.

vinckama: I was not sure of how to do this, this push me to find a workaround in https://github.com/DataDog/dd-sensitive-data-scanner/pull/88/commits/2125da14fcf39f78732226636b22bf197175b845

fuchsnj (approved):

sds/src/secondary_validation/nir_checksum.rs
vinckama: It's literaly the number 97

From [wikipedia page](https://fr.wikipedia.org/wiki/Num%C3%A9ro_de_s%C3%A9curit%C3%A9_sociale_en_France) (in French):

> clé de contrôle = complément à 97 du nombre formé par les 13 premiers chiffres du NIR modulo 97 
(complément au NIR pour la Sécurité sociale)

translation
> control key = complement to 97 of the number formed by the first 13 digits of the NIR modulo 97 
(complement to NIR for Social Security)

PR #89, opened 2024-07-30
Added tests for exclude keywords rule

StephenWakely:
APR-199

This adds a couple of tests to test that the exclude keywords rule will cause the rule not to match. Another test checks the behaviour when both include and exclude rules are specified.

### DIFF
--- a/sds-go/go/scanner_test.go
+++ b/sds-go/go/scanner_test.go
@@ -574,6 +574,18 @@ func TestExclude(t *testing.T) {
 	defer scanner.Delete()
 
 	testData := map[string]testResult{
+		"this is a potato 4111 1111 1111 1111": {
+			mutated: true,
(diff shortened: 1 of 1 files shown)


sds-go/go/scanner_test.go
```diff
@@ -558,6 +558,87 @@ func TestPartialRedactEnd(t *testing.T) {
...
+	}
+
+	scanner, err := CreateScanner(rules)
+	if err != nil {
+		t.Fatal("failed to create the scanner:", err.Error())
+	}
+	defer scanner.Delete()
+
+	testData := map[string]testResult{
+		"this is a potato 4111 1111 1111 1111": {
```
remeh, line 615: Can you add this result case in the unit test above as well? It is interesting to have it there as well 👍

remeh (approved):

PR #90, opened 2024-07-30
Make rules in go binding more generic

fuchsnj:
This adjusts the go bindings to make the `Rule` more generic so implementations of Rules outside of this repo can still be used with the go bindings (instead of just the built-in regex rules). A `RegexConfig` interface was created for any rule that is able to create a `Rule` (which is  a native pointer to the actual rule). Internally instead of just serializing a JSON list of configs, each rule is now created independently, added to a list, and then a scanner is created with that list. The user-experience is very similar to before, but now you can create arbitrary rules to create any rule type that SDS supports.

Other changes:
- The go package name was changed from `sds` to `dd_sds` to more closely match the naming conventions of this repo


sds-go/go/scanner.go
```diff
@@ -1,9 +1,9 @@
-package sds
+package dd_sds
 
 import (
 	"bytes"
 	"encoding/binary"
-	"encoding/json"
+	//"encoding/json"
```
vinckama, line 6:
```suggestion
```

sds-go/go/dd_sds.h
```diff
@@ -3,11 +3,19 @@
...
 const char* scan(long scanner_id, const void* event, long event_size, long *retsize, long *retcap, const char** error);
 
 void free_vec(const char* string, long len, long cap);
 void free_string(const char* string);
+
+void free_any_rule(long rule_ptr);
+
+long create_rule_list();
+void append_rule_to_list(long rule_ptr, long list_ptr);
+void free_rule_list(long list_ptr);
```
vinckama, line 21: If I read it correctly, all this logic is hidden from the library interface, and this is really good to keep our interface simple 👍

vinckama (approved):
LGTM with a minor comment. 🎉

sds-go/go/dd_sds.h
fuchsnj: Correct, when you create a `Scanner`, all of this is done automatically. You just pass a list of "rule configs", and the rules will be allocated / added to the list automatically for you

trosenblatt:
This looks good to me. One caveat though is that Rule has been changed to RuleConfig which will make all existing go bindings fail when they are going to bump the dd-sensitive-data-scanner

PR #91, opened 2024-08-06
Blocked rules in scan function

trosenblatt:
[JIRA_LINK](https://datadoghq.atlassian.net/browse/SDS-421)
This enabled the shared library to accept list of blocked rules index in each scan. Those rules are excluded from the scan.


sds/src/scanner/mod.rs
```diff
@@ -163,7 +163,11 @@ impl Scanner {
         ScannerBuilder::new(rules)
     }
 
-    pub fn scan<E: Event>(&self, event: &mut E) -> Vec<RuleMatch> {
+    pub fn scan<E: Event>(
+        &self,
+        event: &mut E,
+        blocked_rules_idx: Option<Vec<usize>>,
```
vinckama, line 169: It does not seem to have difference of behaviour between `None` and `Some(vec![])`. In that case, the `None` state does not add more information than an empty list. To simplify our interface, I suggest to drop the `Option` and keep only the `Vec<usize>`. It will reduce the indirection of the parameter.

sds/src/scanner/mod.rs
```diff
@@ -163,7 +163,11 @@ impl Scanner {
         ScannerBuilder::new(rules)
     }
 
-    pub fn scan<E: Event>(&self, event: &mut E) -> Vec<RuleMatch> {
+    pub fn scan<E: Event>(
```
vinckama, line 166: As we start to have more arguments, I believe it's a good opportunity to add a rustdoc documenting the goal of the two arguments

sds/src/scanner/mod.rs
```diff
@@ -457,6 +471,9 @@ impl<'a, E: Encoding> ContentVisitor<'a> for ScannerContentVisitor<'a, E> {
         let mut group_data: AHashMap<TypeId, Box<dyn Any>> = AHashMap::new();
 
         rule_visitor.visit_rule_indices(|rule_index| {
+            if self.blocked_rules.contains(&rule_index) {
```
vinckama, line 474: For each rule and each leaf of the event, this method will be called. It could be expensive for big events.
I think it can be optimised a bit if this is pushed to the `RuleIndexVisitor`, `ActiveRuleTree` to drop the rules earlier in the scan pipeline. I'm not sure it's possible but I wonder if we can perform the exclusion of rule once per event and not one per leaf. Did you explore this option?

vinckama (commented):
I added 3 comments. I'm not sure of the performance of the exclusion (described in the last comment).

sds/src/scanner/mod.rs
trosenblatt: faire enough

sds/src/scanner/mod.rs
trosenblatt: ok

sds/src/scanner/mod.rs
trosenblatt: Yes this is a tricky part. 
This is not super optimal as we will hash every rule_index visited for each content. 
Alternative maybe what you suggested but not sure how to do that, or faster structure search (like bool array of excluded rules)

trosenblatt: I will dig into it a bit more

trosenblatt: I create a Vec version so that inference is O(1) and there is no hashing involved. Not sure how to do better using the ScopedRuleset. Maybe @fuchsnj you have a better idea ?

fuchsnj: A `Vec` is fine as long as the list stays small (<10 or so), after that we probably need something else. The `ScopedRuleset` wouldn't support this very well, since the "excluded" rules in there are still scanned for Multipass V0.

Does the `blocked_rules` really change frequently enough to need to pass it into _every_ scan function? Can we just re-build the scanner with the new rules whenever this list changes?

fuchsnj: The perf of the `Vec` is _probably_ fine. It seems like this will usually be an empty `Vec`, so the `.contains()` will be fast. It's only populated during a slow rollout, so that should be somewhat temporary.

sds/src/scanner/mod.rs
fuchsnj: A `Vec` doesn't allocate memory until something it pushed into it, so it should be fine to use an empty `Vec` instead of `None`. (Although this is usually passed in as a slice instead of `Vec` to be more generic, so `blocked_rules_idx: &[usize]`)

sds/src/scanner/mod.rs
fuchsnj: An alternative to re-building the entire scanner, we could add a function to the scanner to disable rules, which could re-build just the `ScopedRuleset` (instead of having to re-compile all of the regexes).

trosenblatt: We can't rebuild the scanner since the slow rollout has a rand part in it, meaning it can change between 2 scans and so we would have to rebuild the scanner very often. 
This is not a contains anymore but a Vec<bool> indexed by rule_idx containing true when the rule is blocked so access is always fast whatever the number of blocked rules. Downside is that it takes more memory for potentially nothing.

fuchsnj: Okay, the `Vec` is probably fine. I would just make sure that the common case of not having a rollout at all doesn't have any significant overhead (don't allocate an array of `false`) for each scan.

Just brainstorming here, but an alternative may be to have multiple `ScopedRuleset`'s, and you can dynamically choose which one to use for each scan. The `Vec` is probably fine though and we can look at perf improvements if they are needed.

trosenblatt: Ok I move to a pure vector + contains search.
When the vector is populated it will be small and transient.

sds/src/scanner/mod.rs
```diff
@@ -439,6 +449,8 @@ struct ScannerContentVisitor<'a, E: Encoding> {
     scanner: &'a Scanner,
     caches: CachePoolGuard<'a>,
     rule_matches: &'a mut Vec<(crate::Path<'static>, Vec<InternalRuleMatch<E>>)>,
+    // Rules that shall be skipped for this scan
+    blocked_rules: &'a Vec<bool>,
```
fuchsnj: Consider making this optional, so the allocation isn't required in the most frequent case (all false). This might not make a significant different either way though since it's only once per scan.

trosenblatt: I moved back to a Vec<usize> containing the list of rules. This is simple and should not impact memory nor CPU in most cases.

vinckama (approved):

PR #92, opened 2024-08-06
make multipass v0 ordering independent (attempt 3)

fuchsnj:
This adds the change to make multipass V0 order independent yet again. It was previously reverted due to an unrelated high-priority bugfix.

### History
- https://github.com/DataDog/dd-sensitive-data-scanner/pull/10
- https://github.com/DataDog/dd-sensitive-data-scanner/pull/35
- https://github.com/DataDog/dd-sensitive-data-scanner/pull/79
- https://github.com/DataDog/dd-sensitive-data-scanner/pull/81
- This PR


sds/src/scanner/mod.rs
```diff
@@ -1500,11 +1542,11 @@ mod test {
 
         // Due to the ordering of the scan (alphabetical in this case), the match from "a-match" is not
         // excluded yet, but "z-match" is, so only 1 match is found.
```
vinckama, line 1544: I think the first part of the comment can be removed, (the updated comment say the contrary of this old part)
```suggestion
```

vinckama (approved):
Approved with a minor comment

trosenblatt:
@fuchsnj just in case you don't remember. The tricky part was in logs-backend as tests were failing (that's the reason why we rolled it back)

PR #93, opened 2024-08-14
Add option to disable multipass v0

fuchsnj:
An option is added to the scanner builder to disable "Multipass V0". Previously this was always on, and now it defaults to on, but can be disabled for the entire scanner.

### DIFF
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -174,10 +174,21 @@ where
     }
 }
 
-#[derive(Default, Debug, PartialEq)]
+#[derive(Debug, PartialEq)]
 struct ScannerFeatures {
     pub should_keywords_match_event_paths: bool,
     pub add_implicit_index_wildcards: bool,
+    pub multipass_v0_enabled: bool,
+}
+
+impl Default for ScannerFeatures {
+    fn default() -> Self {
+        Self {
+            should_keywords_match_event_paths: false,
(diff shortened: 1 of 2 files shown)


sds/src/scanner/mod.rs
```diff
@@ -186,6 +186,7 @@ pub struct Scanner {
     cache_pool: CachePool,
     scanner_features: ScannerFeatures,
     metrics: ScannerMetrics,
+    multipass_v0_enabled: bool,
```
vinckama, line 189: I think we want to have this part of `ScannerFeatures`, so avoid having a giant list of attributes for configs

fuchsnj: Sure, I can change it. The `ScannerFeatures` were mostly useful before we had the `ScannerBuilder` since you had to pass in all options. Now the new feature is set by the builder, so you don't pass it in at all unless you want to change it from the default. (This change won't change anything use facing)

fuchsnj: https://github.com/DataDog/dd-sensitive-data-scanner/pull/93/commits/93c975d3298f26a6adc789dd8c0fbee102a450d0

vinckama: Agree, For bindings this should simplify a bit the code by forwarding a byte array of ScannerFeatures instead of having to forward each argument independently.

vinckama (approved):

PR #95, opened 2024-08-20
🎨 [Included keywords matching path improvement 1/N] Make active_tree_count a struct called NodeCounter

artslidd:
## Description

- Make a new data structure called `NodeCounter` that contains the count for the current `active_tree_count`, and the upcoming `true_positive_rule_idx` vector that will contain the rule indices that have matched at the current path. (👈  This PR)

(Next PRs 👇)
- Create another vector (that acts as a stack) that contains the list of sanitized segments until that point. The sanitized path would be the concatenation of all the vector, linked using the unified linked char
- Create a sanitize method on the segment struct
- Run the path matching every time we push a segment on all the rules that have not yet matched the path (looking at the `true_positive_rule_idx`
- Store the result in the `true_positive_rule_idx`
- When scanning a leaf, retrieve the information if the rule is a true positive or not, and either include it or exclude it

### DIFF
--- a/sds/src/scoped_ruleset/mod.rs
+++ b/sds/src/scoped_ruleset/mod.rs
@@ -68,7 +68,11 @@ impl ScopedRuleSet {
                 rule_tree: &self.tree,
                 index_wildcard_match: false,
             }],
-            active_tree_count: vec![1],
+            true_positive_rule_idx: vec![],
+            active_node_counter: vec![NodeCounter {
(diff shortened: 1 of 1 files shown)


fuchsnj (approved):
This looks fine. I'll wait until more of this is implemented to suggest any optimizations (e.g. combining multiple `Vec`s or something similar) to see what makes sense

PR #96, opened 2024-08-22
🎨 [Included keywords matching path improvement 2/N] Sanitize segment and push it to a new list when pushing a new segment

artslidd:
## Description

**Create another vector (that acts as a stack) that contains the list of sanitized segments until that point. The sanitized path would be the concatenation of all the vector, linked using the unified linked char (👈  this PR)
Create a sanitize method on the segment struct  (👈 this PR)**

(Next PRs 👇)

Run the path matching every time we push a segment on all the rules that have not yet matched the path (looking at the true_positive_rule_idx
When scanning a leaf, retrieve the information if the rule is a true positive or not, and either include it or exclude it

### DIFF
--- a/sds/src/proximity_keywords/mod.rs
+++ b/sds/src/proximity_keywords/mod.rs
@@ -15,6 +15,7 @@ use regex_syntax::ast::{
     Group, GroupKind, Literal, LiteralKind, Position, Span,
 };
 use thiserror::Error;
+use crate::proximity_keywords::BypassStandardizePathResult::{ShouldBypassAndAllLowercase, ShouldBypassAndAllUppercase, ShouldNotBypass};
 
 const MAX_KEYWORD_COUNT: usize = 50;
 pub const MAX_LOOK_AHEAD_CHARACTER_COUNT: usize = 50;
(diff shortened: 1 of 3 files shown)


sds/src/scoped_ruleset/mod.rs
```diff
@@ -190,6 +195,10 @@ where
                 });
             }
         }
+
+        // Sanitize the segment and push it
+        self.sanitized_segments_until_node.push(segment.sanitize());
```
fuchsnj, line 200: You are pushing to `sanitized_segments_until_node`, but never popping. I think this needs a `.pop` in `.pop_segment`?

artslidd: Yes my bad, I'm adding the pop part

artslidd: [Pop sanitized_segments_until_node when popping a segment](https://github.com/DataDog/dd-sensitive-data-scanner/pull/96/commits/61bd7347f48efd240d84f60924c60d70ba4febc2)

sds/src/proximity_keywords/mod.rs
```diff
@@ -276,26 +278,37 @@ fn get_char_type(c: &char) -> CharType {
...
         let is_upper = char.is_ascii_uppercase();
         let is_lower = char.is_ascii_lowercase();
         // If it's neither an uppercase character nor a lowercase character, return false
         if !is_lower && !is_upper {
-            return false;
+            return BypassStandardizePathResult::NoBypass;
         }
         all_lower = all_lower && is_lower;
         all_upper = all_upper && is_upper;
         // If we realise that we don't have all uppercase nor all lowercase, return false
```
fuchsnj, line 300: The comments are a bit out of date here

fuchsnj (approved):

sds/src/proximity_keywords/mod.rs
artslidd: Ha yes good catch!

PR #97, opened 2024-08-22
🎨 [Included keywords matching path improvement 3/N] Sanitize segment and push it to a new list when pushing a new segment

artslidd:
## Description

Create another vector (that acts as a stack) that contains the list of sanitized segments until that point. The sanitized path would be the concatenation of all the vector, linked using the unified linked char (☝️ previous PRs)
Create a sanitize method on the segment struct  (☝️ previous PRs)

**Introduce `find_true_positive_rules_from_current_path` method on ContentVisitor, that pushes rule indices onto the current `true_positive_rule_idx`** (👈 this PR)

(Next PRs 👇)

Run the path matching every time we push a segment on all the rules that have not yet matched the path (looking at the true_positive_rule_idx
When scanning a leaf, retrieve the information if the rule is a true positive or not, and either include it or exclude it

### DIFF
--- a/sds/src/scoped_ruleset/mod.rs
+++ b/sds/src/scoped_ruleset/mod.rs
@@ -214,20 +214,34 @@ where
         // Sanitize the segment and push it
         self.sanitized_segments_until_node.push(segment.sanitize());
 
+        let true_positive_rules_count = if self.should_keywords_match_event_paths {
+            self.content_visitor
+                .find_true_positive_rules_from_current_path(
+                    self.sanitized_segments_until_node.as_slice(),
+                    &mut self.true_positive_rule_idx,
+                )
+        } else {
+            0
+        };
+
         // The new number of active trees is the number of new trees pushed
         self.active_node_counter.push(NodeCounter {
             active_tree_count: self.tree_nodes.len() - tree_nodes_len,
-            true_positive_rules_count: 0,
+            true_positive_rules_count,
         });
 
         self.path.segments.push(segment);
     }
 
     fn pop_segment(&mut self) {
-        let num_active_trees = self.active_node_counter.pop().unwrap().active_tree_count;
-        for _ in 0..num_active_trees {
+        let node_counter = self.active_node_counter.pop().unwrap();
+        for _ in 0..node_counter.active_tree_count {
             // The rules from the last node are no longer active, so remove them.
             let _popped = self.tree_nodes.pop();
+        }
+        for _ in 0..node_counter.true_positive_rules_count {
+            // The true positive rule indices from the last node are no longer active, remove them.
+            let _popped = self.true_positive_rule_idx.pop();
         }
         self.sanitized_segments_until_node.pop();
         self.path.segments.pop();
--- a/sds/src/scoped_ruleset/mod.rs
+++ b/sds/src/scoped_ruleset/mod.rs
@@ -122,6 +122,12 @@ pub trait ContentVisitor<'path> {
         rules: RuleIndexVisitor,
         is_excluded: ExclusionCheck<'content_visitor>,
     ) -> bool;
+
+    fn find_true_positive_rules_from_current_path(
+        &self,
+        sanitized_segments: &[Cow<str>],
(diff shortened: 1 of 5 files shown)


sds/src/scoped_ruleset/mod.rs
```diff
@@ -199,22 +215,52 @@ where
...
+                    sanitized_segment
+                        .as_ref()
+                        .map_or(None::<Cow<str>>, |x| Some(x.clone()))
+                })
+                .fold(String::new(), |mut a, b| {
+                    a.reserve(b.len() + 1);
+                    a.push_str(&*b);
+                    a.push(UNIFIED_LINK_CHAR);
+                    a
+                });
```
artslidd, line 232: This technique is more efficient than collect: https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=bacb488f2c7df299c244403556cd9b25

sds/src/scoped_ruleset/mod.rs
```diff
@@ -199,21 +215,53 @@ where
...
+                .sanitized_segments_until_node
+                .iter()
+                .filter_map(|sanitized_segment| {
+                    sanitized_segment.as_ref().map(|seg| Some(seg.clone()))
+                })
+                .fold(String::new(), |mut a, b| {
+                    let b_str = b.expect(
+                        "In the filter_map above, we make sure to filter out the None variants",
+                    );
+                    a.reserve(b_str.len() + 1);
```
fuchsnj, line 229: `String::reserve` is quite expensive. This is what actually does the allocation, so it would be good to minimize this as much as possible. Luckily, we already have a list of all of the segments, so you can just do a quick sum over the size of the segments to get the total length, and create a string of the correct capacity with only a single allocation with `String::with_capacity`.

I went ahead and implemented this to make sure it's actually better, and I'm seeing a decent perf improvement

```rust
let mut total_len: usize = self
                .sanitized_segments_until_node
                .iter()
                .flatten()
                .map(|x| x.len() + 1)
                .sum();
            if total_len > 0 {
                // subtract the len of the last link char that isn't needed
                total_len -= 1;
            }

            let mut current_sanitized_path = String::with_capacity(total_len);
            for (i, segment) in self
                .sanitized_segments_until_node
                .iter()
                .cloned()
                .flatten()
                .enumerate()
            {
                if i != 0 {
                    current_sanitized_path.push(UNIFIED_LINK_CHAR);
                }
                current_sanitized_path.push_str(segment.as_ref());
            }
```

image

fuchsnj (commented):
Suggested some improvements

sds/src/scoped_ruleset/mod.rs
```diff
@@ -199,21 +215,53 @@ where
         // Sanitize the segment and push it
         self.sanitized_segments_until_node.push(segment.sanitize());
 
+        let true_positive_rules_count = if self.should_keywords_match_event_paths {
+            let mut current_sanitized_path = self
+                .sanitized_segments_until_node
+                .iter()
+                .filter_map(|sanitized_segment| {
```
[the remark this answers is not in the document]
artslidd: I agree, this is a bit convoluted and inefficient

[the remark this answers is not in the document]
artslidd: Your way of doing is better thx!

sds/src/scoped_ruleset/mod.rs
artslidd: [Improve current_sanitized_path creation algorithm](https://github.com/DataDog/dd-sensitive-data-scanner/pull/97/commits/6f75d45d277ec11cb4631c5a9e8eb7aec29c94e1)

artslidd:
@fuchsnj I added a commit: [Add better comments regarding sanitized_segments_until_node being a V…](https://github.com/DataDog/dd-sensitive-data-scanner/pull/97/commits/5b369b0f1c9ecaef5ca30fbc9c01051bbfa1d61f) for the Vec<Option> that was not totally obvious

fuchsnj (approved):

PR #98, opened 2024-08-23
🎨 [Included keywords matching path improvement 4/4] When scanning a leaf, retrieve information if the rule was a true positive

artslidd:
## Description

Create another vector (that acts as a stack) that contains the list of sanitized segments until that point. The sanitized path would be the concatenation of all the vector, linked using the unified linked char (☝️ previous PRs)
Create a sanitize method on the segment struct  (☝️ previous PRs)

Introduce `find_true_positive_rules_from_current_path` method on ContentVisitor, that pushes rule indices onto the current `true_positive_rule_idx`

⬇️  This PR

**Run the path matching every time we push a segment on all the rules that have not yet matched the path (looking at the true_positive_rule_idx
When scanning a leaf, retrieve the information if the rule is a true positive or not, and either include it or exclude it**

### DIFF
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -72,7 +72,7 @@ pub trait CompiledRuleDyn: Send + Sync {
         exclusion_check: &ExclusionCheck<'_>,
         excluded_matches: &mut AHashSet<String>,
         match_emitter: &mut dyn MatchEmitter,
-        should_keywords_match_event_paths: bool,
+        true_positive_rule_idx: &[usize],
         scanner_labels: &Labels,
     );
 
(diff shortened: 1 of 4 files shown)


sds/src/scanner/mod.rs
```diff
@@ -60,19 +60,20 @@ where
 pub trait CompiledRuleDyn: Send + Sync {
     fn get_match_action(&self) -> &MatchAction;
     fn get_scope(&self) -> &Scope;
-    fn get_included_keywords(&self) -> Option<&CompiledIncludedProximityKeywords>;
+    fn get_included_keywords(&self) -> Option<&CompiledIncludedProximityKeywords> {
+        None
+    }
```
artslidd, line 65: Thanks @fuchsnj, didn't know we could implement defaults in traits!

PR #101, opened 2024-08-30
Add global stats

fuchsnj:
[jira ticket](https://datadoghq.atlassian.net/browse/SDS-460)

This adds global metrics, mostly around scanner creation / rules to better understand memory usage of scanners.
These metrics are no-op unless a recorder is explicitly set for the `metrics` crate.

There are 2 additional metrics that would be useful, but were not implemented due to complexity / cost.

- Regex size. This is very expensive to calculate since it's not directly accessible. It tends to be significantly lower than the cache size used for the regex, so it's not as important. This memory usage also isn't multiplied by the parallelism like the regex cache is.
- Actual parallelism per scanner/rule. This is very complicated to calculate without impacting performance. It can be roughly estimated by comparing the total memory usage by the number of scanners * avg cache size of scanners, which is probably enough to have a good understanding of this. Some new changes are coming soon to consolidate rules across Scanners which would significantly change how this is implemented anyway and this can be revisited then.

### DIFF
--- a/sds-bindings-utils/Cargo.lock
+++ b/sds-bindings-utils/Cargo.lock
@@ -174,6 +174,7 @@ dependencies = [
  "crc32fast",
  "farmhash",
  "iban_validate",
+ "lazy_static",
  "metrics",
  "metrics-util",
  "nom",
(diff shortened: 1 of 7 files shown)


trosenblatt (approved):
LGTM !

PR #104, opened 2024-09-06
Enable the `bench` feature automatically for benchmarks

fuchsnj:
I just recently learned of a "trick" to enable features for the current crate in tests/benches only, which is used here to automatically enable the `bench` feature for benchmarks so it doesn't need to be done manually anymore. Some code only existed to make sure this feature was enabled, which can now be removed.


trosenblatt (approved):
LGTM

PR #107, opened 2024-09-25
Match Validation: Add HTTP Custom support

trosenblatt:
[JIRA](https://datadoghq.atlassian.net/browse/SDS-479)

2/3 split of this https://github.com/DataDog/dd-sensitive-data-scanner/pull/105

HTTP Custom match implementation

### DIFF
--- a/sds/src/match_validation/http_validator.rs
+++ b/sds/src/match_validation/http_validator.rs
@@ -1,0 +1,20 @@
+use super::match_validator::MatchValidator;
+use crate::{CompiledRuleDyn, RuleMatch};
+use async_trait::async_trait;
+
+pub struct HttpValidator {
+    // Otherwise struct is not allocated by compiler opt
+    // will be removed in next PR
+    _useless: bool,
+}
+
+impl HttpValidator {
+    pub fn new() -> Self {
+        HttpValidator { _useless: false }
+    }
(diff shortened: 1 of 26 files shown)


sds/src/match_validation/http_validator.rs
```diff
@@ -1,20 +1,201 @@
...
+        builder
     }
 }
 
 #[async_trait]
 impl MatchValidator for HttpValidator {
-    async fn validate(&self, _: &mut Vec<RuleMatch>, _: &Vec<Box<dyn CompiledRuleDyn>>) {}
+    async fn validate(&self, matches: &mut Vec<RuleMatch>, _: &Vec<Box<dyn CompiledRuleDyn>>) {
+        // Let's reqwest the HTTP API endpoint to validate the matches
+        let client = Client::new();
```
fuchsnj, line 133: This should be able to re-use the same global `Client` that is used in the AWS validator.

fuchsnj (approved):

sds/src/match_validation/http_validator.rs
trosenblatt: true will do the same

trosenblatt: FIXED

PR #110, opened 2024-09-27
Match Validation: Support for AWS Validator

trosenblatt:
[JIRA](https://datadoghq.atlassian.net/browse/SDS-479)


sds/src/match_validation/aws_validator.rs
```diff
@@ -1,22 +1,145 @@
...
+                Some(MatchValidationType::Aws(AwsType::AwsSecret(_))) => {
+                    aws_secret_matches_idx.push(idx);
+                }
+                _ => {}
+            }
+        }
+
+        let mut match_status_per_pairs_of_matches_idx: AHashMap<(usize, usize), MatchStatus> =
+            AHashMap::new();
+        for id_index in 0..aws_id_matches_idx.len() {
```
fuchsnj, line 52: You should avoid directly using the index unless it's needed. In this case you only use the values from the array, so you can directly iterate over the values instead (`for id in aws_id_matches_idx`)

sds/src/match_validation/aws_validator.rs
```diff
@@ -1,22 +1,145 @@
...
+                    (
+                        aws_id_matches_idx[id_index],
+                        aws_secret_matches_idx[secret_index],
+                    ),
+                    MatchStatus::NotChecked,
+                );
+            }
+        }
+
+        let client = Client::new();
```
fuchsnj, line 64: Caching this and re-using it globally can improve performance. It's thread-safe, and will be able to re-use existing connections / caches / thread-pools etc.

sds/src/match_validation/aws_validator.rs
```diff
@@ -1,22 +1,145 @@
...
+        // Let's try all combination of aws_id and aws_secret
+        let futures = match_status_per_pairs_of_matches_idx.iter_mut().map(
+            |((id_index, secret_index), match_status)| {
+                let client = client.clone();
+                let match_id = &matches[*id_index];
+                let match_secret = &matches[*secret_index];
+                async move {
```
fuchsnj, line 71: The `async` stuff is fine here, just wanted to let you know that `reqwest` does have a sync variant (different import) if for some reason we find the async is to much work / not enough benefit.

sds/src/match_validation/aws_validator.rs
```diff
@@ -1,22 +1,145 @@
...
+                        .post(self.config.aws_sts_endpoint.as_str())
+                        .headers(headers)
+                        .body(body)
+                        .timeout(self.config.timeout)
+                        .send()
+                        .await;
+
+                    match res {
+                        Ok(val) => {
+                            // If status is 200-299, then it's valid we can safely update the match status
```
fuchsnj, line 93: It seems like you are making this more generic later (validators can pick ranges that are considered "success"?) but just wanted to make sure you knew that the `StatusCode` type has methods to do this for you if it helps at all: https://docs.rs/reqwest/0.12.7/reqwest/struct.StatusCode.html#method.is_success

sds/src/match_validation/validator_utils.rs
```diff
@@ -8,11 +10,47 @@ use super::{
...
+    );
+    headers.insert("host", "sts.us-east-1.amazonaws.com".parse().unwrap());
+
+    let body = "Action=GetCallerIdentity&Version=2011-06-15".to_string();
+    let s = aws_sign_v4::AwsSign::new(
+        "POST",
+        endpoint,
+        &datetime,
+        &headers,
+        "us-east-1", // default region
```
fuchsnj, line 46: Does this need to be customizable, or is `us-east-1` always valid?

sds/src/scanner/mod.rs
```diff
@@ -2525,6 +2527,203 @@ mod test {
             _ => assert!(false),
         }
     }
+    #[cfg(feature = "match_validation")]
+    #[tokio::test]
```
fuchsnj, line 2531: I'm glad you found the `#[tokio::test]` which is very helpful here. Just keep in mind that by default this creates a single-threaded tokio runtime.  It may not be obvious what's going on if you didn't know that. The singled-threaded one is fine most of the time, but if you ever find a test that seems to deadlock / stop progress and you can't figure out why, it's usually because of that.
To switch, use `#[tokio::test(flavor = "multi_thread")]`

sds/src/match_validation/aws_validator.rs
trosenblatt: yes this is correct and I fixed it in this pr thks to clippy 
https://github.com/DataDog/dd-sensitive-data-scanner/pull/111

sds/src/match_validation/aws_validator.rs
trosenblatt: True I will use it for aws

sds/src/match_validation/aws_validator.rs
trosenblatt: true I will create a lazy_static!

sds/src/match_validation/aws_validator.rs
```diff
@@ -1,22 +1,145 @@
...
+        matches: &mut Vec<RuleMatch>,
+        scanner_rules: &Vec<Box<dyn CompiledRuleDyn>>,
+    ) {
+        // Let's regroup matches per type
+        let mut aws_id_matches_idx = vec![];
+        let mut aws_secret_matches_idx = vec![];
+
+        for (idx, m) in matches.iter().enumerate() {
+            let rule: &Box<dyn CompiledRuleDyn> = &scanner_rules[m.rule_index];
+            match rule.get_match_validation_type() {
```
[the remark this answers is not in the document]
trosenblatt: good call I just fixed it

sds/src/match_validation/aws_validator.rs
trosenblatt: done

sds/src/match_validation/validator_utils.rs
trosenblatt: We are using a default sts endpoint that is hosted in us-east-1 
It should solved all request from all endpoint. But this is a good point I haven't tested with credentials from other DC. I will do that as a follow up

sds/src/scanner/mod.rs
trosenblatt: good to know thanks

sds/src/match_validation/validator_utils.rs
trosenblatt: We are relying on a global sts endpoint that is hosted in us-east-1
All IAM account are global so I guess since that works with the test credential it should be good

PR #114, opened 2024-10-02
Reduce memory usage by sharing regexes

fuchsnj:
[Jira](https://datadoghq.atlassian.net/browse/SDS-459)

Currently, each scanner instance owns their own copy of each regex used. If there are many scanners that use the same regex, this creates unnecessary regex copies and can consume quite a bit of memory. Some of the memory wasted comes from the compiled regex itself, but most of it comes from the regex [cache](https://docs.rs/regex-automata/0.4.8/regex_automata/meta/struct.Cache.html) that is needed.

This PR changes that, and creates a global regex store. Only a single regex will ever exist, no matter how many scanners are using it.

### Regex Cache strategy

By default, the [regex](https://crates.io/crates/regex) crate automatically handles the regex cache. Each regex has an internal (highly optimized) pool to lend out caches as they are needed. The number of caches stored per regex is roughly equivalent to the number of concurrent threads using a regex at the same time. Even though this pool is highly optimized already, due to the design you have to acquire a cache for every use of a regex. A single `.scan()` call in SDS can run a regex scan _thousands_ of times, consuming a significant amount of CPU just for obtaining the cache.

To improve on the default behavior, previously a change was made to manually manage the regex caches. Each scanner would keep a pool of a set of caches (using the same pool implementation as the `regex` crate), and you would only need to acquire this set of caches a single time per `.scan()` call. This significantly reduced CPU usage in highly concurrent uses of SDS, and is the current implementation.

As mentioned previously, _most_ of the memory usage is actually from these regex caches, so keeping separate copies for every scanner can't stay, a new regex caching strategy is needed.

This PR also introduces a global regex _cache_ pool. A few different implementations were tried, but ultimately the pool from the `regex` crate still ended up being the best option. For each `.scan()`, a `RegexCaches` is requested from the global cache pool. `RegexCaches` holds 1 cache for _every_ regex used in SDS. The caches in a `RegexCaches` are lazily created as needed. They are also evicted when a regex is no longer used (e.g. The last `Scanner` using the regex is dropped).
This is accomplished by using a [SlotMap](https://docs.rs/slotmap/1.0.7/slotmap/struct.SlotMap.html) in the global regex store to store a mapping of a`SlotMap` key to a (weak) reference to each shared regex. Then in each `RegexCaches`, the caches are stored in a [SecondaryMap](https://docs.rs/slotmap/1.0.7/slotmap/struct.SecondaryMap.html) using the same key from the `SlotMap`. The `SlotMap` is occasionally garbage-collected to find any dead weak references. The `SecondaryMap`'s are automatically cleaned up due to how keys are managed by the `SlotMap`, old caches are simply overwritten when new ones are created. (this is described in more detail [here](https://docs.rs/slotmap/1.0.7/slotmap/struct.SecondaryMap.html))

It is quite easy to use the new caching strategy, so its use was extended to the regexes used for included keywords (which was still using the default `regex` pool). This _significantly_ improved performance for regex rules using included keywords, as you can see in the benchmarks below.

### Benchmarks

Although this PR is focusing on reducing memory usage, this required a change in the caching strategy which can significantly impact performance. The goal was to introduce this change without any significant performance regression.

New benchmarks were added simulating a high-thread contention workload (and also single-threaded variants to compare to). These benchmarks measure the amount of time required to scan 13 KB of logs with 31 made-up regex patterns 256 times (~3.3MB total). The "single-string" tests scan each line of the logs (108 lines) individually, and the "large-event" tests use a single event, which is a map where each field is a line from the log. The multi-threaded tests use a thread-pool with 32 threads.

Overall, the performance has actually improved. Performance has greatly improved for rules using included keywords (due to them now using the better caching strategy), and is mostly neutral for other regex rules.

Multi-threaded benchmarks
image

Single-threaded benchmarks
image

### Metrics

- `scanner.regex_cache_size` was removed. It measured the sum of regex caches for a single scanner. This doesn't make as much sense with the changes in this PR, and the way it was implemented was mostly useless, so it was removed.

- `scanner.total_regex_cache_size` was added, which is the sum of 1 cache for _every_ (shared) regex in the SDS library.

- `scanner.total_regexes` was added, which is the total number of shared regexes in the SDS library.


sds/src/scanner/mod.rs
```diff
@@ -261,16 +249,19 @@ impl Scanner {
...
-            event,
-            ScannerContentVisitor {
-                scanner: self,
-                caches,
-                rule_matches: &mut rule_matches_list,
-                blocked_rules: &blocked_rules_idx,
-                excluded_matches: &mut excluded_matches,
-            },
-        );
+        access_regex_caches(|regex_caches| {
```
trosenblatt, line 252: FYI I will very soon work on a scan cache (for ML Rules) so this will be renamed pretty quickly :D

sds/src/scanner/regex_rule/regex_cache_store.rs
```diff
@@ -0,0 +1,45 @@
...
+
+pub fn access_regex_caches<T>(func: impl FnOnce(&mut RegexCaches) -> T) -> T {
+    // This function isn't strictly necessary, but it makes it easier to change the implementation
+    // later
+    let mut caches = REGEX_CACHE_STORE.get();
+    func(&mut caches)
+}
+
+pub struct RegexCaches {
+    map: SecondaryMap<RegexCacheKey, regex_automata::meta::Cache>,
```
trosenblatt, line 21: for my own understanding I don't get the main difference between secondaryMap and AHashMap what the benefit of SecondaryMap here ?

sds/src/scanner/mod.rs
fuchsnj: This only works for regexes, so I'm not sure why it needs to be renamed. If ML Rules wants to cache a regex, this will still work :)

sds/src/scanner/regex_rule/regex_cache_store.rs
fuchsnj: You can read more about it here: https://docs.rs/slotmap/latest/slotmap/secondary/struct.SecondaryMap.html

The difference is quite important and it comes down to the implementation. Internally it's actually just wrapping a `Vec`, so lookups are faster (no hashing required). The main benefit is due to how the keys are managed. Once the primary `SlotMap` has a key removed, new keys can re-use the same index in the `Vec` so the `SecondaryMap` will automatically clean up memory from unused cached without having to manually remove old entries from every `SecondaryMap` manually.

sds/src/scanner/mod.rs
trosenblatt: no but we want to cache other things

sds/src/scanner/regex_rule/regex_cache_store.rs
trosenblatt: ok I was missing the SlotMap part but it now makes sense thks

PR #116, opened 2024-10-04
Removing openssl dependency to rely on rust-tls

trosenblatt:
To fix processing issue relying on libssl1.1

### DIFF
--- a/sds/Cargo.toml
+++ b/sds/Cargo.toml
@@ -40,7 +40,7 @@ serde_yaml = "0.9.34"
 lazy_static = "1.5.0"
 futures = "0.3.30"
 async-std = "1.13.0"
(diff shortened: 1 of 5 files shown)


sds/Cargo.toml
```diff
@@ -40,7 +40,7 @@ serde_yaml = "0.9.34"
 lazy_static = "1.5.0"
 futures = "0.3.30"
 async-std = "1.13.0"
-reqwest = "0.12"
+reqwest = { version = "0.12", default-features = false, features = ["charset", "http2", "macos-system-configuration", "rustls-tls"] }
```
fuchsnj, line 43: You can probably just add `rustls-tls`, but this is fine as a quick fix until we can verify that is the same

fuchsnj (approved):

PR #128, opened 2024-10-21
Bump go staticcheck the previous version was crashing ...

trosenblatt:
I believe the action we are using has updated its go version and it is no longer compatible with our staticcheck

Added Go 1.23 support 
Staticcheck 2024.1 has full support for iterators / range-over-func. Furthermore, SA1015 will skip any code targeting Go 1.23 or newer, as it is now possible to use time.Tick without leaking memory.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,7 @@ name: Test Suite
 on:
   pull_request:
   merge_group:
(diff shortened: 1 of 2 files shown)


vinckama (approved):
LGTM, please add a description explaining why this is fixing the issue

PR #129, opened 2024-10-21
Support CustomHttp with several endpoint (eu, us, etc)

trosenblatt:
[JIRA](https://datadoghq.atlassian.net/browse/SDS-382)

This PR comes from a comment made by @vinckama about how we would support CustomHttp with multiple domain such   as datadog with api1, eu1, us1, us3, us5 etc ....
The basic idea here is that we would add a pattern in the endpoint **$HOST** that would be replaced by all the hosts in the hosts items
i.e. If we had
endpoint: https://$HOST/api/v1/validate
hosts:
  - api.datadoghq.com
  - api.datadoghq.eu
  
We would try the match on https://api.datadoghq.com and https://api.datadoghq.eu

### DIFF
--- a/sds/src/match_validation/config.rs
+++ b/sds/src/match_validation/config.rs
@@ -74,17 +74,27 @@ pub struct HttpValidatorConfig {
 }
 
 impl HttpValidatorConfig {
-    pub fn new(endpoint: &str, hosts: Vec<String>) -> Self {
-        // Replace $HOSTS in endpoint and build the endpoints vector
+    pub fn new(endpoint: &str, hosts: Vec<String>) -> Result<Self, String> {
+        // Handle errors cases
+        // - endpoint contains $HOST but no hosts are provided
+        // - endpoint does not contain $HOST but hosts are provided
+        if endpoint.contains("$HOST") && hosts.is_empty() {
+            return Err("Endpoint contains $HOST but no hosts are provided".to_string());
+        }
+        if !endpoint.contains("$HOST") && !hosts.is_empty() {
+            return Err("Endpoint does not contain $HOST but hosts are provided".to_string());
+        }
+
+        // Replace $HOST in endpoint and build the endpoints vector
         let mut endpoints = vec![];
         for host in hosts {
-            endpoints.push(endpoint.replace("$HOSTS", &host));
+            endpoints.push(endpoint.replace("$HOST", &host));
         }
         if endpoints.is_empty() {
             // If no hosts are provided, use the endpoint as is
             endpoints.push(endpoint.to_string());
         }
-        HttpValidatorConfig {
+        Ok(HttpValidatorConfig {
             endpoints,
             method: HttpMethod::Get,
             request_header: vec![],
@@ -95,7 +105,7 @@ impl HttpValidatorConfig {
             options: HttpValidatorOption {
                 timeout: Duration::from_secs(DEFAULT_HTTPS_TIMEOUT_SEC),
             },
-        }
+        })
     }
 }
 
@@ -138,19 +148,39 @@ mod tests {
 
     #[test]
     fn test_http_validator_config_no_hosts() {
-        let config = HttpValidatorConfig::new("http://localhost/test", vec![]);
+        let config = HttpValidatorConfig::new("http://localhost/test", vec![]).unwrap();
         assert_eq!(config.endpoints, vec!["http://localhost/test"]);
     }
 
     #[test]
     fn test_http_validator_config_with_hosts() {
         let config = HttpValidatorConfig::new(
-            "http://localhost/$HOSTS",
+            "http://localhost/$HOST",
             vec!["us".to_string(), "eu".to_string()],
-        );
+        )
+        .unwrap();
         assert_eq!(
             config.endpoints,
             vec!["http://localhost/us", "http://localhost/eu"]
+        );
+    }
+
+    #[test]
+    fn test_http_validator_config_error_cases() {
+        let config = HttpValidatorConfig::new("http://localhost/$HOST", vec![]).unwrap_err();
+        assert_eq!(
+            config,
+            "Endpoint contains $HOST but no hosts are provided".to_string()
+        );
+    }
+
+    #[test]
+    fn test_http_validator_config_error_cases_with_hosts() {
+        let config =
+            HttpValidatorConfig::new("http://localhost/test", vec!["us".to_string()]).unwrap_err();
+        assert_eq!(
+            config,
+            "Endpoint does not contain $HOST but hosts are provided".to_string()
         );
     }
 }
--- a/sds/src/match_validation/config.rs
+++ b/sds/src/match_validation/config.rs
@@ -207,15 +207,17 @@ impl MatchValidationType {
             }
         }
     }
-    pub fn into_match_validator(&self) -> Box<dyn MatchValidator> {
+    pub fn into_match_validator(&self) -> Result<Box<dyn MatchValidator>, String> {
         match self {
             MatchValidationType::Aws(aws_type) => match aws_type {
-                AwsType::AwsSecret(aws_config) => Box::new(AwsValidator::new(aws_config.clone())),
-                _ => panic!("This aws type shall not be used to create a validator"),
+                AwsType::AwsSecret(aws_config) => {
+                    Ok(Box::new(AwsValidator::new(aws_config.clone())))
+                }
+                _ => Err("This aws type shall not be used to create a validator".to_string()),
             },
-            MatchValidationType::CustomHttp(http_config) => {
-                Box::new(HttpValidator::new_from_config(http_config.clone()))
-            }
+            MatchValidationType::CustomHttp(http_config) => Ok(Box::new(
+                HttpValidator::new_from_config(http_config.clone()),
+            )),
         }
     }
 }
--- a/sds/src/match_validation/config.rs
+++ b/sds/src/match_validation/config.rs
@@ -74,7 +74,7 @@ pub struct HttpValidatorConfig {
 }
 
 impl HttpValidatorConfig {
-    pub fn new(endpoint: &str, hosts: Vec<String>) -> Result<Self, String> {
+    fn new(endpoint: &str, hosts: Vec<String>) -> Result<Self, String> {
         // Handle errors cases
         // - endpoint contains $HOST but no hosts are provided
         // - endpoint does not contain $HOST but hosts are provided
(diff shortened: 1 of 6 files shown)


sds/src/match_validation/config.rs
```diff
@@ -120,5 +129,28 @@ impl MatchValidationType {
...
+    fn test_http_validator_config_no_hosts() {
+        let config = HttpValidatorConfig::new("http://localhost/test", vec![]);
+        assert_eq!(config.endpoints, vec!["http://localhost/test"]);
+    }
+
+    #[test]
+    fn test_http_validator_config_with_hosts() {
+        let config = HttpValidatorConfig::new(
+            "http://localhost/$HOSTS",
+            vec!["us".to_string(), "eu".to_string()],
```
vinckama, line 149: I think we wait to raise an error is an host is specified in the endpoint but no host is available in the list and   same if there are endpoints but `$HOSTS` is missing

sds/src/match_validation/config.rs
```diff
@@ -74,9 +74,18 @@ pub struct HttpValidatorConfig {
 }
 
 impl HttpValidatorConfig {
-    pub fn new(endpoint: &str) -> Self {
+    pub fn new(endpoint: &str, hosts: Vec<String>) -> Self {
+        // Replace $HOSTS in endpoint and build the endpoints vector
+        let mut endpoints = vec![];
+        for host in hosts {
+            endpoints.push(endpoint.replace("$HOSTS", &host));
```
vinckama, line 81: From a pure API point of view, I would expect to write `$HOST` and not `$HOSTS` because in the endpoint, I expect a single host to replace `$HOST`

sds/src/match_validation/http_validator.rs
```diff
@@ -136,71 +152,86 @@ impl HttpValidatorHelper {
...
-                        request_builder = HTTP_CLIENT.patch(&self.config.endpoint);
+        // let's build a map of match status per endpoint and per match_idx
+        let mut match_status_per_endpoint_and_match = AHashMap::new();
+        for (idx, _) in matches.iter().enumerate() {
+            for endpoint in &self.config.endpoints {
+                match_status_per_endpoint_and_match
+                    .entry((idx, endpoint))
+                    .or_insert_with(|| MatchStatus::NotChecked);
+            }
+        }
```
vinckama, line 163: You can possibly use map/flatmap/collect methods from the iterator interface to directly create the map with the right capacity and populated items:
```suggestion
        let mut match_status_per_endpoint_and_match: AHashMap<_,_> =
       matches.iter().enumerate()
           .flat_map(|(idx, _)| {
               self.config.endpoints.iter()
                   .map(move |endpoint| ((idx, endpoint), MatchStatus::NotChecked))
        }).collect();
```

sds/src/match_validation/http_validator.rs
```diff
@@ -136,71 +152,86 @@ impl HttpValidatorHelper {
...
+                        }
+                        HttpMethod::Put => {
+                            request_builder = HTTP_CLIENT.put(*endpoint);
+                        }
+                        HttpMethod::Delete => {
+                            request_builder = HTTP_CLIENT.delete(*endpoint);
+                        }
+                        HttpMethod::Patch => {
+                            request_builder = HTTP_CLIENT.patch(*endpoint);
+                        }
```
vinckama, line 185: Minor, I think a more rust idiomatic way to write this code is to assign the variable directly:
```suggestion
                    let mut request_builder= match self.config.method {
                        HttpMethod::Get => HTTP_CLIENT.get(*endpoint),
                        HttpMethod::Post =>  HTTP_CLIENT.post(*endpoint),
                        HttpMethod::Put =>  HTTP_CLIENT.put(*endpoint),
                        HttpMethod::Delete => HTTP_CLIENT.delete(*endpoint),
                        HttpMethod::Patch => HTTP_CLIENT.patch(*endpoint)
                    };
```

sds/src/scanner/mod.rs
```diff
@@ -2559,6 +2563,44 @@ mod test {
         assert_eq!(matches[1].match_status, MatchStatus::Valid);
         assert_eq!(matches[2].match_status, MatchStatus::Valid);
     }
+
+    #[tokio::test]
+    async fn test_mock_endpoint_with_multiple_hosts() {
```
vinckama, line 2568: I think we also want to tests a few other cases. In particular, it would be interesting to validate that:
- if all status are invalid, then the result is invalid. 
- if one status is valid and one is erroring, then the result is valid
- If one status is erroring and all others are invalid, then the result is erroring (unchecked), [because we don't know if the erroring status validate the match or invalidate it]

vinckama (commented):
I added a few comments and suggestions

sds/src/match_validation/config.rs
```diff
@@ -74,9 +74,18 @@ pub struct HttpValidatorConfig {
...
+        }
+        if endpoints.is_empty() {
+            // If no hosts are provided, use the endpoint as is
+            endpoints.push(endpoint.to_string());
+        }
         HttpValidatorConfig {
-            endpoint: endpoint.to_string(),
+            endpoints,
```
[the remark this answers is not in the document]
trosenblatt: good idea

sds/src/match_validation/http_validator.rs
trosenblatt: very true

sds/src/match_validation/config.rs
```diff
@@ -120,5 +129,28 @@ impl MatchValidationType {
 #[derive(PartialEq, Eq, Hash)]
 pub enum InternalMatchValidationType {
     Aws,
-    CustomHttp(String),
+    CustomHttp(Vec<String>),
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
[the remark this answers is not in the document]
trosenblatt: why not added it

sds/src/match_validation/http_validator.rs
```diff
@@ -136,71 +152,86 @@ impl HttpValidatorHelper {
...
+                            *match_status = MatchStatus::Error(fmt::format(format_args!(
+                                "Unexpected HTTP status code {}",
+                                val.status().as_u16()
+                            )));
+                        }
+                        Err(err) => {
+                            // TODO(trosenblatt) emit a metrics for this
+                            *match_status = MatchStatus::Error(fmt::format(format_args!(
+                                "Error making HTTP request: {}",
+                                err
```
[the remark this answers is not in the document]
trosenblatt: Why would they be returned ?

sds/src/match_validation/config.rs
trosenblatt: done

sds/src/scanner/mod.rs
trosenblatt: this is already tested in the match_status.rs via the merge function.
I don't think we should cover all cases here the idea is to confirm that the merge is called. 
merge is already unit tested

sds/src/match_validation/http_validator.rs
```diff
@@ -136,71 +154,102 @@ impl HttpValidatorHelper {
...
-                        for valid_range in &self.config.valid_http_status_code {
-                            if valid_range.contains(&val.status().as_u16()) {
-                                m.match_status = MatchStatus::Valid;
-                                return;
+        let futures = match_status_per_endpoint_and_match.iter_mut().map(
+            |((match_idx, endpoint), match_status)| {
+                let match_value = matches[*match_idx].match_value.as_ref().unwrap();
+                async move {
+                    let mut request_builder: reqwest::RequestBuilder;
+                    request_builder = match self.config.method {
```
vinckama, line 174: I don't think there is any conflict with `RequestBuilder` struct name, so there is not need to explicitly import it from  `reqwest`. Line 173 could be `let mut request_builder: RequestBuilder;`. Even better, you can assign directly the value to `request_builder`
```suggestion
                    let mut request_builder = match self.config.method {
```

sds/src/match_validation/config.rs
```diff
@@ -74,9 +74,28 @@ pub struct HttpValidatorConfig {
 }
 
 impl HttpValidatorConfig {
-    pub fn new(endpoint: &str) -> Self {
-        HttpValidatorConfig {
-            endpoint: endpoint.to_string(),
+    pub fn new(endpoint: &str, hosts: Vec<String>) -> Result<Self, String> {
```
vinckama, line 77: As there is a builder now, I think it is better is to force everyone to use it. TO do so, we can make this method private. This will require to move the `HttpValidatorConfigBuilder` in this class, but I think it's for the best as the builder will be closer to the actual object built (`HttpValidatorConfig`)
```suggestion
   fn new(endpoint: &str, hosts: Vec<String>) -> Result<Self, String> {
```

sds/src/match_validation/http_validator.rs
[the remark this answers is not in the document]
trosenblatt: Nope this is an error generated by the reqwest lib
- no connection to host
- timeout 
- ssl issue
- so on so forth, I don't think there is any risk that a query param will be leaked here

sds/src/match_validation/http_validator.rs
```diff
@@ -18,13 +19,16 @@ pub struct HttpValidator {
 }
 
 impl HttpValidator {
-    pub fn new(config: HttpValidatorConfig) -> Self {
-        HttpValidator { config }
+    pub fn new_clone_config(config: &HttpValidatorConfig) -> Self {
+        HttpValidator {
+            config: config.clone(),
+        }
```
vinckama, line 25: Instead of cloning the class in the `new` method, I think it is more idiomatic to keep the keep the ownership of the config attribute in this class and to push the clone to the in the caller.

sds/src/match_validation/validator_utils.rs
```diff
@@ -14,7 +14,7 @@ pub fn new_match_validator_from_type(config: &MatchValidationType) -> Box<dyn Ma
             _ => panic!("This aws type shall not be used to create a validator"),
         },
         MatchValidationType::CustomHttp(http_config) => {
-            Box::new(HttpValidator::new(http_config.clone()))
+            Box::new(HttpValidator::new_clone_config(http_config))
```
vinckama, line 17: I did not realised that in the previous PRs, but this can be moved directly in the `impl MatchValidationType` with a method like `pub fn into_match_validator(self) -> Box<dyn MatchValidator> {`, and so then the code in the scanner will be `match_validation_type.clone().into_match_validator()`. This will reduce the number of imported files. 
It can be tackled in a follow-up PR as it is not directly related to the current change.

vinckama, line 17: I also think that this method should returned a `Result<Box<dyn MatchValidator>, MatchValidationError>` similar to the [`RegexValidationError`](https://github.com/DataDog/dd-sensitive-data-scanner/blob/f0325a5e0eaf2ecdd9b899bf9e7e69a9e7d43895/sds/src/validation.rs#L7) in this [`CreateScannerError`](https://github.com/DataDog/dd-sensitive-data-scanner/blob/bbe8040fae6452d66efe4dec20c957d6f3e794a4/sds/src/scanner/error.rs#L20). In fact, we don't want the scanner code to panic but rather to return errors to the user.

Same as for the previous comment, it can be tackled in a follow-up PR.

vinckama (commented):
I added a few more comments and suggestion, let me know what you think

sds/src/match_validation/validator_utils.rs
trosenblatt: I disagree here, the new_clone_config will never panic it just clones the configuration data

trosenblatt: I fixed that in this PR it is fair enouh

vinckama: I mean, I think the method `new_match_validator_from_type` should return a result instead of panic

vinckama (approved):
LGTM, there is a pending comment but it can definitely be tackled in a follow-up PR

sds/src/match_validation/validator_utils.rs
trosenblatt: Oh my bad I misunderstood your point. I agree will add a Result

trosenblatt: I did it let me know what you think

PR #130, opened 2024-10-22
Use rustls-tls-native-roots instead of rustls-native

trosenblatt:
From slack
Bruce Guenter
  9:22 PM
A recent PR to dd-sensitive-data-scanner switched the reqwest crate to use rustls. However, the default feature flag for that pulls in webpki-roots. That crate is the first MPL crate that is actually compiled into OPW, which uses SDS, and we would like to avoid that license complication at this point. Additionally, the README for webpki-roots [says](https://github.com/rustls/webpki-roots?tab=readme-ov-file#warning):
For applications that are deployed to end-users and cannot be recompiled, or which need certification before deployment, consider a library that uses the platform native certificate verifier
Based on that, would you be open to switching to [the ](https://docs.rs/reqwest/latest/reqwest/#optional-features)rustls-tls-native-roots[ feature flag on ](https://docs.rs/reqwest/latest/reqwest/#optional-features)reqwest, which drops that dependency? If so, I'll PR it. (edited


artslidd (approved):
LGTM, you might need to re-run the license script

PR #131, opened 2024-10-22
Match Validation: Extract AWS Secret access key from the match to support current and future aws rule

trosenblatt:
[JIRA](https://datadoghq.atlassian.net/browse/SDS-548)
Current AWS Secret access key standard rule extract more than the secret.
This is done in order to reduce false positive. 

It first detect the presence of a char like this [^A-Za-z0-9/+]

This PR aims to extract the secret only before using it to create AWS signature in the futures.

Ideally this should be fixed in the regex itself using capture groups, but this is not implemented yet. In order to unlock match validation feature I implemented here. 
It should work even when the capture will be implemented

### DIFF
--- a/sds/src/match_validation/aws_validator.rs
+++ b/sds/src/match_validation/aws_validator.rs
@@ -16,6 +16,9 @@ use super::{
 
 lazy_static! {
     static ref AWS_CLIENT: Client = Client::new();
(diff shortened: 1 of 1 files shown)


fuchsnj (approved):
Approving as a temporary solution until capture groups are supported

PR #133, opened 2024-10-28
New validation error types

fbryden:
These validation error types are to be used in the migrated tests in the sds shared library.

[JIRA LINK](https://datadoghq.atlassian.net/browse/SDS-341)

### DIFF
--- a/sds-go/go/scanner.go
+++ b/sds-go/go/scanner.go
@@ -80,7 +80,9 @@ func CreateScanner(ruleConfigs []RuleConfig) (*Scanner, error) {
 			return nil, ErrInvalidKeywords
 		case -4: // rust: CreateScannerError::InvalidMatchAction
 			return nil, ErrInvalidMatchAction
-		case -5: // rust panic
+		case -5: // rust: InvalidMetadata::
+			return nil, ErrInvalidMatchAction
+		case -6: // rust panic
 			if errorString != nil {
 				defer C.free_string(errorString)
 				return nil, fmt.Errorf("internal panic: %v", C.GoString(errorString))
--- a/sds-go/go/scanner.go
+++ b/sds-go/go/scanner.go
@@ -20,7 +20,6 @@ var (
 	ErrInvalidRegex       error = fmt.Errorf("invalid regex")
 	ErrInvalidKeywords    error = fmt.Errorf("invalid keywords")
 	ErrInvalidMatchAction error = fmt.Errorf("invalid match action")
-	ErrInvalidMetadata    error = fmt.Errorf("invalid metadata")
 )
 
 // Scanner wraps an SDS scanner.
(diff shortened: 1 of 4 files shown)


sds/src/scanner/error.rs
```diff
@@ -45,6 +49,55 @@ pub enum MatchValidationError {
     NoMatchValidationType,
 }
 
+#[derive(Debug, PartialEq, Eq, Error)]
+pub enum ScannerMetadataError {
```
vinckama, line 53: As these errors are not used in the dd-sensitive-data-scanner code. I wonder if it would be possible to move it to internal library with a different error type. It would help to split the ownership between internal library and public engine.

sds-go/go/scanner.go
```diff
@@ -80,7 +81,9 @@ func CreateScanner(ruleConfigs []RuleConfig) (*Scanner, error) {
 			return nil, ErrInvalidKeywords
 		case -4: // rust: CreateScannerError::InvalidMatchAction
 			return nil, ErrInvalidMatchAction
-		case -5: // rust panic
+		case -5: // rust: CreateScannerError::InvalidMetadata
```
vinckama, line 84: I don't really understand why we need to expose this new error type in the Go Binding. In fact, the validation of standard rules should not have much impact on the scanner itself.

vinckama (commented):
I added two general comments on the approach

sds/src/proximity_keywords/mod.rs
```diff
@@ -439,9 +439,21 @@ pub enum ProximityKeywordsValidationError {
...
+
+    #[error("`{0}` keyword contains uppercase characters - matching is case-insensitive so using lower cased characters yields the same results")]
+    KeywordHasUppercase(String),
+
     #[error("Keywords cannot be longer than the look ahead character count ({0})")]
     KeywordTooLong(usize),
 
+    #[error("`{0}` keyword contains invalid characters - only alphanumeric characters and the following ones (` `, `@`, `.`, `\\`, `/`, `#`, `&`, `+`, `-`)")]
+    KeywordInvalidCharacters(String),
+
```
fbryden, line 456: @vinckama I am migrating the rest of the code to our internal library (sds-shared) - however I feel like this part is relevant to this library, as it is simply an extension of the coverage of `ProximityKeywordsValidationError`. This error is already in the library and had some corner cases it was lacking.

The core library uses it in `compile_keywords_to_ast`. Rules are compiled and the errors are generated when malformed rules are given. Perhaps the coverage of that function could be improved at some point and these new errors would be used there?

sds/src/lib.rs
```diff
@@ -38,7 +38,8 @@ pub use path::{Path, PathSegment};
 pub use rule_match::{ReplacementType, RuleMatch};
 pub use scanner::shared_pool::{SharedPool, SharedPoolGuard};
 
-pub use scanner::error::MatchValidationError;
+pub use proximity_keywords::ProximityKeywordsValidationError;
+pub use scanner::error::{MatchValidationError, ScannerMetadataError};
```
fbryden, line 42: And these changes would also be required if I still wanted to use the public library's errors

sds/src/proximity_keywords/mod.rs
fbryden: Well turns out one can create rules in this package that violate these limitations described by these errors without any issue. I'll thus remove them from here and close the PR.

fbryden:
Closing this PR as all the changes will be made in the internal repo.

PR #139, opened 2024-11-13
Add `gofmt` & `go mod tidy` in CI

vinckama:
Add `gofmt` and `go mod tidy` in the CI to make sure go files are always formatted. Before fixing formatting, the [CI failed](https://github.com/DataDog/dd-sensitive-data-scanner/actions/runs/11818928119/job/32927650824), it demonstrates that the formatter works as expected:

Screenshot 2024-11-13 at 15 08 56

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,6 +57,9 @@ jobs:
           go install honnef.co/go/tools/cmd/staticcheck@v0.5.1
           export PATH=$PATH:$(go env GOPATH)/bin
           LD_LIBRARY_PATH=../rust/target/release staticcheck ./...
+          gofmt -s -w . && git diff --exit-code
+          # Replace the gofmt command above by `gofmt -s -d .` when the following is resolved https://github.com/golang/go/issues/46289
+          go mod tidy -diff
       - name: "Run the Go unit tests"
         run: |
           cd go
--- a/sds-go/go/regex_rule.go
+++ b/sds-go/go/regex_rule.go
@@ -8,8 +8,8 @@ import "C"
 
 import (
 	"encoding/json"
-	"unsafe"
 	"fmt"
+	"unsafe"
 )
 
 type RegexRuleConfig struct {


.github/workflows/test.yml
```diff
@@ -57,6 +57,8 @@ jobs:
           go install honnef.co/go/tools/cmd/staticcheck@v0.5.1
           export PATH=$PATH:$(go env GOPATH)/bin
           LD_LIBRARY_PATH=../rust/target/release staticcheck ./...
+          gofmt -s -w . && git diff --exit-code
```
fbryden, line 60: I'm not exactly sure how this works in a CI...but `-s` will "fix" the code. Is `-d` the more "suitable" option here, which displays the issues rather than fixes them?

vinckama: I think the challenge here is that the failed command is github do not make the test to fail. See this [discussion](https://github.com/golang/go/issues/46289#issuecomment-862490917), I think it is still not implemented yet

vinckama: I tried on [1e49fd0](https://github.com/DataDog/dd-sensitive-data-scanner/pull/139/commits/1e49fd04ee3ad966b7d279031318fb9186e7b53d), to double check the behavior, moving the PR to draft again

fbryden: ah I was going to suggest something I thought was smart - but I just found it in the thread you linked lol

fbryden: I'm currently testing this in my own PR:
```
  script:
    # Only the sds-go folder needs to be checked since generation-checks takes care of the generated code
    - output=$(gofmt -d -s sds-go)
    - if [[ -n "$output" ]]; then echo "gofmt failed"; echo "$output"; exit 1; fi
```

fbryden: https://gitlab.ddbuild.io/DataDog/sds-shared-library/-/jobs/705361200
This works

vinckama: The script suggested is more complicated than the current one. I'm not sure it's a real improvement

fbryden: Sorry I'd missed your suggestion - I agree with you and I'll be following that in my own PR.

fbryden (approved):
LGTM

PR #140, opened 2024-11-13
Use automata fork

fbryden:
We can configure the MAX_POOL_STACKS at runtime!

### DIFF
--- a/sds/Cargo.toml
+++ b/sds/Cargo.toml
@@ -22,9 +22,10 @@ ahash = "0.8.7"
 # Overriding with fix repo due to: https://github.com/seiflotfy/rust-farmhash/pull/16
 farmhash = { git = "https://github.com/fuchsnj/rust-farmhash", rev = "82d80b689d65fbd378b13deff10cdd07794df64e" }
 nom = "7.1.3"
-regex = { git = "https://github.com/fbryden/regex" }
+# regex = { git = "https://github.com/fbryden/regex", tag = "1.9.5", rev = "554469b8c1116322f3c0a054ceeb610224f8ac65" }
+regex = { git = "https://github.com/fbryden/regex", rev = "554469b8c1116322f3c0a054ceeb610224f8ac65" }
 regex-automata = { git = "https://github.com/fbryden/regex", rev = "dccd48f48b6c772cb69eb864a8d95cd27515a53a" }
-regex-syntax = "0.7.5" # { git = "https://github.com/fbryden/regex" }
+regex-syntax = { git = "https://github.com/fbryden/regex", tag = "regex-syntax-0.7.5" }
 serde = { version = "1.0", features = ["derive"] }
 serde_with = "3.6.1"
 thiserror = "1.0.58"
--- a/sds/Cargo.toml
+++ b/sds/Cargo.toml
@@ -23,6 +23,7 @@ ahash = "0.8.7"
 farmhash = { git = "https://github.com/fuchsnj/rust-farmhash", rev = "82d80b689d65fbd378b13deff10cdd07794df64e" }
 nom = "7.1.3"
 regex = "1.9.5"
+regex-automata = "0.4.7"
 # Switch over to the original repo when this issue is resolved: https://github.com/rust-lang/regex/issues/1241
 regex-automata-fork = { git = "https://github.com/fbryden/regex", rev = "6952250af962ca3e364da47382b16dba9c703431", package = "regex-automata" }
 regex-syntax = "0.7.5"
--- a/sds/Cargo.toml
+++ b/sds/Cargo.toml
@@ -23,7 +23,7 @@ ahash = "0.8.7"
 farmhash = { git = "https://github.com/fuchsnj/rust-farmhash", rev = "82d80b689d65fbd378b13deff10cdd07794df64e" }
 nom = "7.1.3"
 regex = { git = "https://github.com/fbryden/regex" }
-regex-automata = { git = "https://github.com/fbryden/regex", rev = "7de87756bbd7c6568d2160206ff7615a5f48e205" }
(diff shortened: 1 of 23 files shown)


sds/src/scanner/regex_rule/regex_cache_store.rs
```diff
@@ -6,8 +6,8 @@ use slotmap::SecondaryMap;
 use std::sync::Arc;
 
 lazy_static! {
-    static ref REGEX_CACHE_STORE: Arc<SharedPool<Box<RegexCaches>>> =
-        Arc::new(SharedPool::new(|| Box::new(RegexCaches::new())));
+    static ref REGEX_CACHE_STORE: Arc<SharedPool<Box<RegexCaches>, 16>> =
+        Arc::new(SharedPool::<_, 16>::new(|| Box::new(RegexCaches::new())));
```
fbryden, line 10: this is the key bit

sds/Cargo.toml
```diff
@@ -22,9 +22,10 @@ ahash = "0.8.7"
 # Overriding with fix repo due to: https://github.com/seiflotfy/rust-farmhash/pull/16
 farmhash = { git = "https://github.com/fuchsnj/rust-farmhash", rev = "82d80b689d65fbd378b13deff10cdd07794df64e" }
 nom = "7.1.3"
-regex = "1.9.5"
-regex-automata = "0.4.4"
-regex-syntax = "0.7.5"
+regex = { git = "https://github.com/fbryden/regex", tag = "1.9.5"}
```
fuchsnj, line 25: We should probably continue using the un-forked `regex` crates for normal regex things (so we continue receiving updates). Only the `pool` needs the forked. You can pull in _just_ `regex-automata` from the forked repo and rename it to something unique to it doesn't conflict with the "real" `regex-automata`. We should also include a link to the issue in the `regex` repo to make this configurable which would allow us to stop using the fork.

sds/src/scanner/shared_pool.rs
```diff
@@ -1,21 +1,74 @@
...
+pub type SharedPoolGuard<'a, T, const MAX_POOL_STACKS: usize> =
+    PoolGuard<'a, T, CachePoolFn<T>, MAX_POOL_STACKS>;
 
 /// This is a simple generic wrapper around `Pool` to make it a bit easier to use
 pub struct SharedPool<T> {
-    pool: Pool<T, CachePoolFn<T>>,
+    pool: Box<dyn MyPoolTrait<T>>,
+}
+
+//MyPoolTrait and MyPoolGuardTrait are used to hide away the constant generic in the Pool
```
fuchsnj, line 14:
```suggestion
// MyPoolTrait and MyPoolGuardTrait are used to hide away the constant generic in the Pool
```

sds/src/scanner/shared_pool.rs
```diff
@@ -1,21 +1,74 @@
...
+    /// greater than or equal to the given count, until 64. Anything greater than 32 will use 64.
+    /// Anything less than or equal to 4 will use 4.
+    ///
+    /// Examples:
+    ///     * new(_, 7) -> 8
+    ///     * new(_, 16) -> 16
+    ///     * new(_, 42) -> 64
+    pub fn new(factory: CachePoolFn<T>, count: usize) -> Self {
+        let pool = match count {
+            x if x <= 4 => Box::new(Pool::<_, _, 4>::new(factory)) as Box<dyn MyPoolTrait<T>>,
```
fuchsnj, line 55: I don't really see the point of using a match statement if every single branch has an if guard. Just use a normal if/else statement at that point to make it more readable.

sds/src/scanner/shared_pool.rs
```diff
@@ -1,21 +1,74 @@
...
+    PoolGuard<'a, T, CachePoolFn<T>, MAX_POOL_STACKS>;
 
 /// This is a simple generic wrapper around `Pool` to make it a bit easier to use
 pub struct SharedPool<T> {
-    pool: Pool<T, CachePoolFn<T>>,
+    pool: Box<dyn MyPoolTrait<T>>,
+}
+
+//MyPoolTrait and MyPoolGuardTrait are used to hide away the constant generic in the Pool
+pub trait MyPoolTrait<T>: Sync + Send {
```
fuchsnj, line 15: 🫣 I thought these were placeholder names for the traits. We can probably think of something a bit more descriptive (although naming things _is_ the hardest part of programming)

fbryden: 🙈🙈🙈🙈
I did see this issue - perhaps I wasn't feeling very inspired last Friday night. 
Anyways I've fixed it now :p

sds/src/scanner/shared_pool.rs
fbryden: Fixed!

sds/src/scanner/regex_rule/regex_cache_store.rs
```diff
@@ -1,24 +1,27 @@
...
     // This function isn't strictly necessary, but it makes it easier to change the implementation
     // later
     let mut caches = REGEX_CACHE_STORE.get();
-    func(&mut caches)
+    func(caches.get_ref())
 }
 
 pub struct RegexCaches {
-    map: SecondaryMap<RegexCacheKey, regex_automata::meta::Cache>,
+    map: SecondaryMap<RegexCacheKey, regex_automata_fork::meta::Cache>,
```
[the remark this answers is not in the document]
fbryden: Unfortunately rust complains in all directions with something like - the names match but the types are distinct - are these different versions of the same crate?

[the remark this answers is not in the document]
fuchsnj: This is because you didn't remove it in enough places. I pushed a [commit](https://github.com/DataDog/dd-sensitive-data-scanner/pull/140/commits/7e74d965074383785e289372e52b8e127d02f8c9) with this change

fuchsnj (approved):

PR #141, opened 2024-11-14
Add support for a group_config object that is created a scanner build time and provided to each get_string_matches 

trosenblatt:
This is largely inspired from the group_data implementation.

The basic idea is that we create the group_config in the `CompiledRuleDyn.process_scanner_config` that is done during scanner build time, so each rule can define their own group and store group specific configuration data 

This group_config is then pass down to the get_string_matches of each CompiledRule so they can have access to it if it exists (there is no GroupConfig for CompiledRegexRule as this is intended for MLRules (they will register in it which feature they need to compute)

[JIRA](https://datadoghq.atlassian.net/browse/SDS-602)

### DIFF
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -71,6 +71,7 @@ pub trait CompiledRuleDyn: Send + Sync {
         content: &str,
         regex_caches: &mut RegexCaches,
         group_data: &mut AHashMap<TypeId, Box<dyn Any>>,
+        group_config: &AHashMap<TypeId, Box<dyn Any + Send + Sync>>,
         exclusion_check: &ExclusionCheck<'_>,
         excluded_matches: &mut AHashSet<String>,
         match_emitter: &mut dyn MatchEmitter,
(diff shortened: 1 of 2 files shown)


sds/src/scanner/mod.rs
```diff
@@ -92,6 +93,11 @@ pub trait CompiledRuleDyn: Send + Sync {
     fn get_match_validation_type(&self) -> Option<&MatchValidationType>;
 
     fn get_internal_match_validation_type(&self) -> Option<&InternalMatchValidationType>;
+
+    fn process_scanner_config(
+        &self,
+        group_config: &mut AHashMap<TypeId, Box<dyn Any + Send + Sync>>,
```
fuchsnj, line 99: You could look into using an `Anymap` here, which is the same abstraction you are writing yourself,
https://crates.io/crates/anymap2

fuchsnj (approved):

sds/src/scanner/mod.rs
trosenblatt: Nice thanks for pointing this out. I will create a follow up PR https://datadoghq.atlassian.net/browse/SDS-612

PR #145, opened 2024-11-19
Implement Event trait for an Hashmap<String, Value> type (for serde_json)

artslidd:
## Description

When you take any event (json), and you deserialize it with serde_json, without knowing the shape upfront, then you end up with an `Hashmap<String, serde_json::Value>`.

I would like to be able to scan this data so I've implemented the Event trait for Hashmap<String, Value>

I use this directly in this other PR: https://github.com/DataDog/sds-shared-library/pull/412

### DIFF
--- a/sds/src/event.rs
+++ b/sds/src/event.rs
@@ -1,7 +1,3 @@
-use core::panic;
-use std::collections::HashMap;
-use std::hash::RandomState;
-
 use crate::encoding::{Encoding, Utf8Encoding};
 use crate::path::Path;
 use crate::PathSegment;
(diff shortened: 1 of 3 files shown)


sds/src/event.rs
```diff
@@ -43,9 +47,90 @@ impl Event for String {
     }
 }
 
+impl Event for serde_json::Value {
```
trosenblatt, line 50: I believe it would make sense to create an event_json.rs instead of pushing everything in the same file no ?

artslidd: [move Event implementation for Hashmap to event_json](https://github.com/DataDog/dd-sensitive-data-scanner/pull/145/commits/335fc5f3496de127c87456a1490fff24274e939b)

sds/src/event_json.rs
```diff
@@ -0,0 +1,182 @@
...
+                VisitOp::Pop,
+                VisitOp::Push(PathSegment::Index(1)),
+                VisitOp::Visit("array".into()),
+                VisitOp::Pop,
+                VisitOp::Pop,
+                VisitOp::Pop,
+            ]
+        );
+
+        let mut leaf = String::new();
```
trosenblatt, line 168: Can you add more comments here ? I don't understand why we are getting leaf == "array".to_string()

trosenblatt:
This looks good to me. I found the test a bit hard to understand some explanation would help the reader understand what and why we are testing this

sds/src/event_json.rs
artslidd: [Add comment to better explain a test](https://github.com/DataDog/dd-sensitive-data-scanner/pull/145/commits/8db4161c844dc576fd753b8a99dd8eba3b69dea8)

artslidd: [Add another comment](https://github.com/DataDog/dd-sensitive-data-scanner/pull/145/commits/1dc6846187202db4f24d013fde9d80e97ef56efb)

trosenblatt (approved):
LGTM ! Thanks for addressing my comments

PR #146, opened 2024-11-19
🐛 Remove feature bench for event_json

artslidd:
This is not needed, and it breaks the benchmarks in the sds-shared-library

### DIFF
--- a/sds/src/lib.rs
+++ b/sds/src/lib.rs
@@ -7,7 +7,6 @@ mod encoding;
 mod event;
 mod match_action;
 
-#[cfg(any(test, feature = "testing", feature = "bench"))]
 mod event_json;
 mod match_validation;
 mod normalization;


sds/src/lib.rs
```diff
@@ -7,7 +7,6 @@ mod encoding;
 mod event;
 mod match_action;
 
-#[cfg(any(test, feature = "testing", feature = "bench"))]
```
fbryden, line 10: We don't want to compile `event_json` in the lib _outside_ of testing and benchmarking - hence this feature flag.

You can pull in this dependency in sds-shared-library by adding the following line to the `[dev-dependencies]` in the Cargo.toml:
```
dd-sds = {workspace = true, features = ["bench"]}
```

artslidd: Actually I was not even sure that we don't want to build this outside of testing and benchmarking. But if this works, I'll go for this option 👍

artslidd:
I'm closing the PR since this works

sds/src/lib.rs
artslidd: Thanks by the way, this works 🙇

PR #153, opened 2024-12-11
Rewrite of contains_excluded_keyword_match

fbryden:
I'll leave the original bounds calculation function + `test_my_get_span_bounds` while you play around with the PR.

- [JIRA link](https://datadoghq.atlassian.net/browse/SDS-648)


sds/src/proximity_keywords/mod.rs
```diff
@@ -527,27 +521,39 @@ mod test {
...
-        assert!(excluded.is_false_positive_match("local!host ping", 11));
-        assert!(!excluded.is_false_positive_match("!hostlocal ping", 11));
-
-        let (_included, excluded) =
-            try_new_compiled_proximity_keyword(30, vec![], vec!["ৎhost".to_string()]).unwrap();
+        let (_included, excluded) = try_new_compiled_proximity_keyword(
+            30,
+            vec![],
+            vec![
+                String::from("host"),
```
fuchsnj, line 528: I think you only need a single excluded keyword to reproduce the issue (e.g. you can delete all but `host`)

fuchsnj (approved):
Approving with 1 small comment

PR #154, opened 2024-12-11
Encode match value for bindings

fuchsnj:
Returned matches need to be encoded for the bindings to be able to use them.


trosenblatt (approved):
LGTM

PR #156, opened 2024-12-20
Validate GitHub tokens with zero padding checksum

vinckama:
Detect valid Github tokens with checksum starting with zeros. Decoding the checksum instead of encoding generated one fixes the padding issue 
Examples in the PR have been generated with [this software](https://github.com/therootcompany/base62-token.js).

https://github.blog/engineering/behind-githubs-new-authentication-token-formats/

> A 32 bit checksum in the last 6 digits of each token strikes the optimal balance between keeping the random token portion at a consistent entropy and enough confidence in the checksum. We start the implementation with a CRC32 algorithm, a standard checksum algorithm. We then encode the result with a Base62 implementation, using leading zeros for padding as needed.

_This is not related to any story in particular._

### DIFF
--- a/sds/src/secondary_validation/github_token_checksum.rs
+++ b/sds/src/secondary_validation/github_token_checksum.rs
@@ -20,11 +20,14 @@ impl Validator for GithubTokenChecksum {
             return false;
         }
 
-        // extract the payload (everything except the last 6 chars)
         let computed_checksum = crc32fast::hash(last_part[..last_part.len() - 6].as_bytes());
-        let computed_checksum_b62 = base62::encode(computed_checksum);
-        // check that the crc is the last 6 chars
-        computed_checksum_b62 == last_part[last_part.len() - 6..]
+        let checksum = base62::decode(last_part[last_part.len() - 6..].as_bytes());
+
+        if checksum.is_err() {
+            return false;
+        }
+
+        computed_checksum as u64 == checksum.unwrap() as u64
     }
 }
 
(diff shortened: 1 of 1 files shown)


sds/src/secondary_validation/github_token_checksum.rs
```diff
@@ -20,11 +20,14 @@ impl Validator for GithubTokenChecksum {
             return false;
         }
 
-        // extract the payload (everything except the last 6 chars)
         let computed_checksum = crc32fast::hash(last_part[..last_part.len() - 6].as_bytes());
-        let computed_checksum_b62 = base62::encode(computed_checksum);
-        // check that the crc is the last 6 chars
-        computed_checksum_b62 == last_part[last_part.len() - 6..]
+        let checksum = base62::decode(last_part[last_part.len() - 6..].as_bytes());
+
```
trosenblatt, line 25: @vinckama it would be great if you can find a non reg test...

vinckama: done in [a6a9439](https://github.com/DataDog/dd-sensitive-data-scanner/pull/156/commits/a6a9439e643fe249e6447696682aa10df6e49820)

trosenblatt (approved):
Well spoted thks for fixing this !

sds/src/secondary_validation/github_token_checksum.rs
artslidd: **suggestion:** 
Would be great here to separate the two parts of the `last_part` &str.

Create one variable called `checksum` and `encoded_checksum` that correspond to `last_part[..n-6]` and `last_part[n-6...]` so that it's clearer that the `last_part` is made of two chunks

‏

PR #157, opened 2025-01-08
Add validation by checksum for Brazilian CPF

fbryden:
Introduce secondary validation by checksum for Brazilian national ID (CPF number).

This was implemented based on the description of the checksum on the [CPF wikipedia page](https://pt.wikipedia.org/wiki/Cadastro_de_Pessoas_F%C3%ADsicas).

PR #158, opened 2025-01-08
Add validation by checksum for Brazilian CNPJ

fbryden:
Introduce secondary validation by checksum for Brazilian national ID (CNPJ number).

This was implemented based on the description of the checksum algorithm on the [CNPJ wikipedia page](https://pt.wikipedia.org/wiki/Cadastro_Nacional_da_Pessoa_Jur%C3%ADdica).

### DIFF
--- a/sds/src/secondary_validation/brazilian_cnpj_checksum.rs
+++ b/sds/src/secondary_validation/brazilian_cnpj_checksum.rs
@@ -1,55 +1,61 @@
+use std::{default, f32::DIGITS};
+
 use crate::secondary_validation::Validator;
+
+use super::get_next_digit;
 
 pub struct BrazilianCnpjChecksum;
 
-const BRAZILIAN_CNPJ_LENGTH: usize = 18;
-// XX.XXX.XXX/YYYY-ZZ
-const BRAZILIAN_CNPJ_SEPARATOR_INDICES: &[usize] = &[2, 6, 10, 15];
+const BRAZILIAN_CNPJ_DIGIT_COUNT: usize = 14;
+
+const BRAZILIAN_CNPJ_V1_MULTIPLIERS: &[u32] = &[5, 4, 3, 2, 9, 8, 7, 6, 5, 4, 3, 2];
+const BRAZILIAN_CNPJ_V2_MULTIPLIERS: &[u32] = &[6, 5, 4, 3, 2, 9, 8, 7, 6, 5, 4, 3];
 
 impl Validator for BrazilianCnpjChecksum {
     // https://pt.wikipedia.org/wiki/Cadastro_Nacional_da_Pessoa_Jur%C3%ADdica
     fn is_valid_match(&self, regex_match: &str) -> bool {
-        // Check if the length of the ID is correct
-        if regex_match.len() != BRAZILIAN_CNPJ_LENGTH {
-            return false;
-        }
-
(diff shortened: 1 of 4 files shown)


sds/src/secondary_validation/brazilian_cnpj_checksum.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl Validator for BrazilianCnpjChecksum {
+    // https://pt.wikipedia.org/wiki/Cadastro_Nacional_da_Pessoa_Jur%C3%ADdica
+    fn is_valid_match(&self, regex_match: &str) -> bool {
+        // Check if the length of the ID is correct
+        if regex_match.len() != BRAZILIAN_CNPJ_LENGTH {
+            return false;
+        }
+
+        // Collect digits from match
```
vinckama, line 17: You can possible avoid to collect the numbers by computing the checksum on the fly. [Here is an example](https://github.com/DataDog/dd-sensitive-data-scanner/blob/9c72d7031f9172c2aa84a1c6a2b0702e53a06880/sds/src/secondary_validation/nhs_check_digit.rs#L14-L30) of checksums where the scanner does not collect the numbers.

vinckama (commented):
I added a suggestion to avoid collecting the numbers

sds/src/secondary_validation/brazilian_cnpj_checksum.rs
```diff
@@ -0,0 +1,84 @@
...
+
+const BRAZILIAN_CNPJ_LENGTH: usize = 18;
+// XX.XXX.XXX/YYYY-ZZ
+const BRAZILIAN_CNPJ_SEPARATOR_INDICES: &[usize] = &[2, 6, 10, 15];
+
+impl Validator for BrazilianCnpjChecksum {
+    // https://pt.wikipedia.org/wiki/Cadastro_Nacional_da_Pessoa_Jur%C3%ADdica
+    fn is_valid_match(&self, regex_match: &str) -> bool {
+        // Check if the length of the ID is correct
+        if regex_match.len() != BRAZILIAN_CNPJ_LENGTH {
```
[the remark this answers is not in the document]
vinckama: also possibly you don't need that at all and you can handle it on the fly. We might want to compute the checksum for an suite of integers composed of 14 numbers. The format of the match should not matter much. It will make the secondary validation a bit less regex dependent. For instance, if we ever want to support accept inputs like: `XXXXXXXX/YYYY-ZZ`, it would be ideal to not have to update the secondary validator.

sds/src/secondary_validation/brazilian_cnpj_checksum.rs
```diff
@@ -0,0 +1,79 @@
...
+        false
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::secondary_validation::*;
+    #[test]
+    fn test_valid_brazilian_cnpj_ids() {
+        let valid_ids = vec!["00.623.904/0001-73", "00.623.904/0001-73"];
```
vinckama, line 56: Same as this [comment](https://github.com/DataDog/dd-sensitive-data-scanner/pull/157#discussion_r1922231381), would it be possible to add a few more examples  to show validated match without separator/with custom separators

vinckama (approved):

PR #159, opened 2025-01-09
Fix panic when included keyword and regex match at the same index

fuchsnj:
A panic could happen if a regex matched at the exact same index as the _start_ of an included keyword. The code was adjusted to only start regex scanning at the _end_ of an included keyword, since those are required to be before a regex match to be valid anyway. This prevent them from overlapping.

### DIFF
--- a/sds/src/scanner/regex_rule/compiled.rs
+++ b/sds/src/scanner/regex_rule/compiled.rs
@@ -131,12 +131,13 @@ impl RegexCompiledRule {
     ) {
         let mut included_keyword_matches = included_keywords.keyword_matches(content);
 
-        'included_keyword_search: while let Some(included_keyword_match_start) =
+        'included_keyword_search: while let Some(included_keyword_match) =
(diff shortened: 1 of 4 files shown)


sds/src/scanner/regex_rule/compiled.rs
```diff
@@ -131,12 +131,12 @@ impl RegexCompiledRule {
...
         let mut included_keyword_matches = included_keywords.keyword_matches(content);
 
-        'included_keyword_search: while let Some(included_keyword_match_start) =
+        'included_keyword_search: while let Some(included_keyword_match) =
             included_keyword_matches.next(regex_caches)
         {
             let true_positive_search = self.true_positive_matches(
                 content,
-                included_keyword_match_start,
+                included_keyword_match.end,
```
fbryden, line 139: This is really nice, such a convenient feature in the end

fbryden (approved):
Thank you for your speedy coding and thinking skills, as always

PR #160, opened 2025-01-23
Remove feature as it's now enabled by default

artslidd:
Remove `should_keywords_match_events_path` as it's now enabled by default, on all orgs, and all DCs


trosenblatt (approved):
LGTM

PR #161, opened 2025-02-06
logs-pattern: Adding wildcard_indices to the scan function

trosenblatt:
Add support for wildcard indices per path in the scanning function (needed by the ML Rules implementation)
[JIRA](https://datadoghq.atlassian.net/browse/SDS-792)


sds/src/scanner/mod.rs
```diff
@@ -300,9 +304,15 @@ impl Scanner {
     // This function scans the given event with the rules configured in the scanner.
     // The event parameter is a mutable reference to the event that should be scanned (implemented the Event trait).
     // The blocked_rules_idx parameter is a list of rule indices that should be skipped for this scan.
+    // The wildcarded_indices parameter is a map containing a list of tuples of (start, end) indices that should be treated as wildcards (for the message key only) per path.
     // this list shall be small (<10), so a linear search is acceptable otherwise performance will be impacted.
     // The return value is a list of RuleMatch objects, which contain information about the matches that were found.
-    pub fn scan<E: Event>(&self, event: &mut E, blocked_rules_idx: Vec<usize>) -> Vec<RuleMatch> {
+    pub fn scan<E: Event>(
```
fuchsnj, line 310: It would be nice if things like  the new `wildcarded_indices` was added to a separate `scan` function. (The `blocked_rules_idx` also). It makes SDS significantly more complicated to understand for new users and requires all uses of SDS to update even if they aren't using this feature (e.g. OPW). The easiest way to get around this is to split this function into 2, one being a simple `scan` with _just_ the event, and another with all the possible options (maybe with a config object or builder or something so unused features don't need to be set).

At a minimum this should at least have a bit more documentation. Unless you are already familiar with this feature, the concept of "wild card indices" for a path doesn't really mean much. What does it actually do? Can this be safely ignored if no ML rules are used?

fuchsnj (approved):
Approving, but I have some suggestions I would like to see potentially as a follow up

sds/src/scanner/mod.rs
trosenblatt: This is a fair point and I completely agree with the comment, especially as I had to rename all scan calls which is pretty painful.
I will create a scanWithOptions method in this PR and restore all previous scan with a single event argument.

trosenblatt: I have done it and will merge this PR. Thks for the review that was a good call !

PR #162, opened 2025-02-14
[SDS-594] Add more tags to serde in order to help the serialization in sds-go

artslidd:
No production changes.

Simply adding serde tags to help with the go serialization

### DIFF
--- a/sds/src/match_validation/config.rs
+++ b/sds/src/match_validation/config.rs
@@ -28,6 +28,7 @@ impl Default for AwsConfig {
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(tag = "kind")]
 pub enum AwsType {
     AwsId,
     AwsSecret(AwsConfig),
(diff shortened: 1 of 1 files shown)


sds/src/match_validation/config.rs
```diff
@@ -28,6 +28,7 @@ impl Default for AwsConfig {
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[serde(tag = "kind")]
```
trosenblatt, line 31: We already have a serializable sds shared version in sds/libs/sds_standard_rules/src/lib.rs. Why do you need to serialize this one exactly ?

trosenblatt (approved):
As discussed LGTM !

sds/src/match_validation/config.rs
artslidd: As discussed in live, I think it will go well with the other structs definition, and also in Golang it will be easier to build the json :)

PR #163, opened 2025-02-19
Add validation for ABA routing transit number

alexandrefouchs:
Add a secondary validation by checksum for ABA routing transit number, based on the on [ABA routing transit number wikipedia page](https://en.wikipedia.org/wiki/ABA_routing_transit_number).

https://datadoghq.atlassian.net/browse/SDS-603


sds/src/secondary_validation/aba_rtn_checksum.rs
```diff
@@ -0,0 +1,62 @@
...
+        for i in 0..9 {
+            if let Some(digit) = get_next_digit(&mut chars) {
+                sums[i % 3] += digit;
+            } else {
+                return false;
+            }
+        }
+
+        let checksum = (3 * sums[0] + 7 * sums[1] + sums[2]) % 10;
+        checksum == 0
```
fbryden, line 24: You could avoid the array allocation and compute the checksum in-place by first defining a statically allocated array of coefficients:
```
static ABA_RNT_MULTIPLIERS: &'static [u32] = &[3, 7, 1];
```

From there you can do something like
```suggestion
        let mut checksum: u32 = 0;

        for i in 0..9 {
            if let Some(digit) = get_next_digit(&mut chars) {
                checksum += ABA_RNT_MULTIPLIERS[i % 3] * digit;
            } else {
                return false;
            }
        }

        checksum % 10 == 0
```

sds/src/secondary_validation/aba_rtn_checksum.rs
```diff
@@ -10,18 +12,22 @@ impl Validator for AbaRtnChecksum {
...
 
-        for i in 0..9 {
+        for _ in 0..9 {
             if let Some(digit) = get_next_digit(&mut chars) {
-                sums[i % 3] += digit;
+                checksum += ABA_RNT_MULTIPLIERS[idx] * digit;
+                idx += 1;
+                if idx == 3 {
+                    idx = 0;
+                } // Avoid %3 operation
```
fbryden, line 24: I'm trying to understand this optimization - what's the theory behind it?

alexandrefouchs: It's probably minor in practice (although, I guess saving allocating a 12 bytes array probably is too) tbh.
The idea was that division in a loop is more costly compared to manually doing it (as we only cycle through the same 3 indexes) which can be better optimized by the Rust compiler IIUC, so may save a tiny bit of CPU.

fbryden (approved):
LGTM

PR #167, opened 2025-02-28
Unify path/content regex behavior

fuchsnj:
[RFC](https://datadoghq.atlassian.net/wiki/spaces/SDS/pages/4792189564/Unifying+path+content+included+keyword+searches)

This unifies the included keyword matching behavior a bit more for path vs content matching.

Path matching allows multiple "word link characters" for multi-word keywords. These are now also added to the content matching regex. If any of the following chars show up in a keyword, _any_ of them are allowed to match in the content itself. They are `-_./<space>`

`_` is also added to the start/end word boundary that is automatically added so `_` can act as a word boundary.

### DIFF
--- a/sds/src/proximity_keywords/included_keywords.rs
+++ b/sds/src/proximity_keywords/included_keywords.rs
@@ -178,4 +178,14 @@ mod test {
             vec![5]
         );
     }
+
+    #[test]
+    fn multi_word_keyword_matches() {
+        let keywords = compile_keywords(30, &["hello world"]);
+
+        assert_eq!(
+            collect_keyword_matches(keywords.keyword_matches("hello-world secret")),
+            vec![0]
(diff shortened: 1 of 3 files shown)


artslidd (approved):
Could we add some more tests to document the new behaviour? Maybe with both content and path matching?

fuchsnj:
@artslidd Added a couple more tests. If you are just looking for documentation, there is a (mini) RFC going over it: https://datadoghq.atlassian.net/wiki/spaces/SDS/pages/4792189564/Unifying+path+content+included+keyword+searches

artslidd:
Also one small feedback if I may, for future PRs. Could you put a link to the ticket in the description so that JIRA is able to link the ticket with the PR and improve the discoverability of it? :)

PR #170, opened 2025-03-11
Make rules more generic

fuchsnj:
To be able to make rules that are composed of other rules, some significant changes were made.

## Shared Data
The `CompiledRule` trait used to have 3 associated types to be able to store shared data, these were
`GroupData`,  `GroupConfig`, and `RuleScanCache`. These were not flexible enough, since it was not feasible to implement a rule that "contained" other rules with different types of shared data. These have been removed, in favor of a new struct that was introduced, `SharedData`.

`SharedData` is a light wrapper around a `Map<TypeId, Any>`, and is now available instead of the associated types. There are different lifetimes that each associated type used, and they were replaced with a `SharedData` matching that lifetime.

`GroupConfig` -> `per_scanner_data`
`GroupData` -> `per_string_data`
`RuleScanCache` -> `per_event_data`

This `SharedData` is significantly more flexible, allowing storing of multiple types dynamically. It also prevents some unnecessary allocations that were being done when shared data wasn't needed.

## CompiledRuleDyn
The main `CompiledRule` trait has been made "dyn compatible" (object safe), which means the `CompiledRuleDyn` trait is no longer needed and was removed 🎉 

## Config

Each configuration currently includes _all_ config fields, even if they aren't specific to the rule type. Common fields were moved out of individual rule configs, and moved to a wrapper type, `RootRuleConfig`, which now contains `match_action`, `scope`, and `match_validation_type`. This not only simplifies adding new rules (don't need to worry about the common fields), but it makes it possible to nest configs.

### DIFF
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -24,6 +24,7 @@ use crate::scanner::config::RuleConfig;
 use crate::scanner::regex_rule::compiled::RegexCompiledRule;
 use crate::scanner::regex_rule::{access_regex_caches, RegexCaches};
 use crate::scanner::scope::Scope;
+pub use crate::scanner::shared_data::SharedData;
 use crate::stats::GLOBAL_STATS;
 use ahash::{AHashMap, AHashSet};
 use regex_automata::Match;
@@ -35,6 +36,7 @@ pub mod error;
 pub mod metrics;
 pub mod regex_rule;
 pub mod scope;
+pub mod shared_data;
 pub mod shared_pool;
 
 #[cfg(test)]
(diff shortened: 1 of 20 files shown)


sds/src/scanner/mod.rs
```diff
@@ -341,17 +319,18 @@ impl ScanOptionBuilder {
...
-    rules: Vec<Box<dyn CompiledRuleDyn>>,
+    rules: Vec<RootCompiledRule>,
     scoped_ruleset: ScopedRuleSet,
     scanner_features: ScannerFeatures,
     metrics: ScannerMetrics,
     labels: Labels,
     match_validators_per_type: AHashMap<InternalMatchValidationType, Box<dyn MatchValidator>>,
-    group_configs: AHashMap<TypeId, Box<dyn Any + Send + Sync>>,
+    per_scanner_data: SharedData,
+    // group_configs: AHashMap<TypeId, Box<dyn Any + Send + Sync>>,
```
trosenblatt, line 329: this can be removed

trosenblatt (approved):
This is great Nathan it simplifies rule configuration and compilation a lot !

trosenblatt (commented):
LGTM I will approved once the main of dd-sds is updated and reflected in this PR

PR #174, opened 2025-04-16
[SDS-1013] Verhoeff algo - luxembourg NIN checksum validator

OBe95:


sds/src/secondary_validation/luxembourg_individual_nin_checksum.rs
```diff
@@ -0,0 +1,90 @@
...
+         * ensuring persons born on the same date have a unique national ID,
+         * then a first check on YYYYMMDDXXX using the Luhn10 algorithm,
+         * and finally a check on YYYYMMDDXXX using the Verhoeff algorithm.
+         */
+        if regex_match.len() != LUXEMBOURG_INDIVIDUALS_NIN_LENGTH {
+            return false;
+        }
+
+        let verhoeff_digit = regex_match.chars().last().unwrap();
+        let luhn_digit = regex_match.chars().nth(regex_match.len() - 2).unwrap();
```
fuchsnj, line 21: Use `https://doc.rust-lang.org/std/iter/trait.DoubleEndedIterator.html#method.nth_back` here, which should be a bit more efficient than checking form the front. As an alternative, you could do an upfront check to ensure the entire match is `ASCII` chars, then do direct indexing into the string (since all ASCII are single-byte).

sds/src/secondary_validation/luxembourg_individual_nin_checksum.rs
```diff
@@ -0,0 +1,90 @@
...
+        let nin_without_checksum = &regex_match[..regex_match.len() - 2];
```
fuchsnj, line 22: Direct string indexing is dangerous and can panic since you haven't verified `regex_match.len() - 2` is a valid char index yet (it could be the middle of a multi-byte character). Either use the `.chars()` iterator, or verify you are working with only ASCII chars first

sds/src/secondary_validation/luxembourg_individual_nin_checksum.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        if !LuhnChecksum.is_valid_match(format!("{}{}", nin_without_checksum, luhn_digit).as_str())
```
fuchsnj, line 24: You should probably stay away from using `format!()` in performance sensitive code. Since you are just concatenating strings there are faster ways to do that. Same for the checksum below.

fuchsnj (approved):

PR #176, opened 2025-04-29
Add test to document ignored rule

vinckama:


fuchsnj:
This is fixed in https://github.com/DataDog/dd-sensitive-data-scanner/compare/fuchsnj/fix_match_validation?expand=1

PR #177, opened 2025-04-29
Fix match validation dropping matches without a validator

fuchsnj:
`validate_matches` would previously only return matches that had an associated "match validator", meaning matches without a validator were removed from the `rule_matches` list.

Rules are now never removed from the list, and instead marked as `NotAvailable` if a match validator is not available for the match.

This fixes the bug found [here](https://github.com/DataDog/dd-sensitive-data-scanner/pull/176)

### DIFF
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -691,6 +691,8 @@ impl ScannerBuilder<'_> {
                         if !match_validators_per_type.contains_key(&internal_type) {
                             match_validators_per_type.insert(internal_type, match_validator);
                             // Let's add return_matches to the scanner features
+                            // TODO Fixme, this implicit behavior could cause issue in case the config is reloaded.
+                            // The scanner features should only be enabled at build time and not based on custom rules.
                             scanner_features.return_matches = true;
                         }
                     } else {
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -431,17 +431,23 @@ impl Scanner {
         }
         // Create MatchValidatorRuleMatch per match_validator_type to pass it to each match_validator
         let mut match_validator_rule_match_per_type = AHashMap::new();
-        for rule_match in rule_matches.drain(..) {
+
+        let mut validated_rule_matches = vec![];
+
+        for mut rule_match in rule_matches.drain(..) {
             let rule = &self.rules[rule_match.rule_index];
             if let Some(match_validation_type) = rule.internal_match_validation_type() {
                 match_validator_rule_match_per_type
                     .entry(match_validation_type)
                     .or_insert_with(Vec::new)
                     .push(rule_match)
+            } else {
+                // There is no match validator for this rule, so mark it as not available.
+                rule_match.match_status.merge(MatchStatus::NotAvailable);
+                validated_rule_matches.push(rule_match);
             }
         }
 
-        // Call the validate per match_validator_type with their matches and the RuleMatch list and collect the results
         match_validator_rule_match_per_type.par_iter_mut().for_each(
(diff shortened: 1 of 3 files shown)


sds/src/scanner/mod.rs
```diff
@@ -691,6 +698,8 @@ impl ScannerBuilder<'_> {
                         if !match_validators_per_type.contains_key(&internal_type) {
                             match_validators_per_type.insert(internal_type, match_validator);
                             // Let's add return_matches to the scanner features
+                            // TODO Fixme, this implicit behavior could cause issue in case the config is reloaded.
+                            // The scanner features should only be enabled at build time and not based on custom rules.
```
trosenblatt, line 702: Jumping on this PR. I don't understand this comment "should only be enabled at build time". 
We are in the build method so why couldn't we enable a scanner feature here ?

trosenblatt (approved):
Thks for fixing this that was a big miss. I added a comment mainly for my understanding

sds/src/scanner/mod.rs
```diff
@@ -455,11 +461,12 @@ impl Scanner {
...
         // Refill the rule_matches with the validated matches
         for (_, mut matches) in match_validator_rule_match_per_type {
-            rule_matches.append(&mut matches);
+            validated_rule_matches.append(&mut matches);
         }
 
         // Sort rule_matches by start index
-        rule_matches.sort_by_key(|rule_match| rule_match.start_index);
+        validated_rule_matches.sort_by_key(|rule_match| rule_match.start_index);
+        *rule_matches = validated_rule_matches;
```
vinckama, line 469: I don't think it's related to this PR but the ordering is different from this [one](https://github.com/DataDog/dd-sensitive-data-scanner/blob/e50ab079ab160ae97f5c2d512641ea16fd575dff/sds/src/scanner/mod.rs#L582-L601). It may be something to align in a follow-up PR to avoid two ordering function

sds/src/scanner/test/match_validation.rs
```diff
@@ -268,6 +268,30 @@ fn test_mock_http_timeout() {
         _ => assert!(false),
     }
 }
+
+#[test]
+fn test_matches_from_rule_without_validation_are_not_ignored() {
```
vinckama, line 273: minor: The name could be more explicit
```suggestion
fn test_matches_from_rule_without_validation_have_not_available_match_validation_status() {
```

vinckama (approved):

sds/src/scanner/mod.rs
fuchsnj: I think this is fine because all overlapping matches have already been removed. Most of the complication is dealing with overlapping matches.

sds/src/scanner/mod.rs
vinckama: Ideally, scanner features should not be enabled silently based on rules, especially when these settings have an impact on third parties (this requires a opened network)

I think it would be better instead to explicitly require `with_return_matches(true)`to enable the feature while creating the scanner. This way, `with_return_matches` will be visible on the user code.

sds/src/scanner/mod.rs
vinckama: I agree that here it's not necessarily impacting our code. However, it looks like a unexpected behavior for users of the library. Someone integrating the library may assume that this method does not impact the order of matches but only alter the `match_status`. I think not alerting the order would make the use of this code less error-prone. The sorting at the end also make the time complexity to at least `O(n*log(n))` where `n` is the number of matches.

Instead of re-ordering the matches, would it be possible to keep the order in the method as so avoid the sorting and avoid the possible re-ordering.

PR #178, opened 2025-04-29
[SDS-966] Add France SSN checksum validator

isabella-garza-datadog:
- This PR adds a checksum validator for the French social security number rule. 
- Related Jira: https://datadoghq.atlassian.net/browse/SDS-966

### DIFF
--- a/sds/src/secondary_validation/france_ssn_checksum.rs
+++ b/sds/src/secondary_validation/france_ssn_checksum.rs
@@ -1,0 +1,92 @@
+use crate::secondary_validation::Validator;
+
+pub struct FranceSsnChecksum;
+
+const FRANCE_SSN_LENGTH: u8 = 15;
+const FRANCE_SSN_CONTROL_KEY_LENGTH: u8 = 2;
+const FRANCE_SSN_FIRST_VALUE_LENGTH: u8 = FRANCE_SSN_LENGTH - FRANCE_SSN_CONTROL_KEY_LENGTH;
+
+impl Validator for FranceSsnChecksum {
+    // https://en.wikipedia.org/wiki/INSEE_code
+    fn is_valid_match(&self, regex_match: &str) -> bool {
+        let mut first_value: i64 = 0;
+        let mut control_key: i64 = 0;
+        let mut count: u8 = 0;
+        let mut penalties: i64 = 0;
+
+        for c in regex_match.chars() {
+            if c.is_ascii_digit() || matches!(c, 'a' | 'A' | 'b' | 'B') {
+                // Calculate the value of the first 13 alphanumeric characters
+                if count < FRANCE_SSN_FIRST_VALUE_LENGTH {
+                    count += 1;
+                    if let Some(digit) = c.to_digit(10) {
+                        first_value = first_value * 10 + digit as i64;
+                    } else {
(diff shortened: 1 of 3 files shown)


sds/src/secondary_validation/france_ssn_checksum.rs
```diff
@@ -0,0 +1,92 @@
...
+
+impl Validator for FranceSsnChecksum {
+    // https://en.wikipedia.org/wiki/INSEE_code
+    fn is_valid_match(&self, regex_match: &str) -> bool {
+        let mut first_value: i64 = 0;
+        let mut control_key: i64 = 0;
+        let mut count: u8 = 0;
+        let mut penalties: i64 = 0;
+
+        for c in regex_match.chars() {
```
fuchsnj, line 17: This loop is doing a lot at once, making it hard to follow the logic / reason about edge cases.

It's currently doing:
- counting valid chars / skipping "invalid" vars
- checking the current count to see what "section" we are in
- checking each individual "section"

If you add a filter to the iterator, it would allow you to split the logic for the "first section" and the "control section" and make this a bit easier to read.

For example:
```rust
let mut valid_chars = regex_match.chars().filter(|c| c.is_ascii_digit() || matches!(c, 'a' | 'A' | 'b' | 'B'));

for c in valid_chars.take(FRANCE_SSN_FIRST_VALUE_LENGTH) {
   // process first section
}

for c in valid_chars.take(FRANCE_SSN_CONTROL_KEY_LENGTH) {
  // process control section
}

if c.next().is_some() {
  // match is too long, return false
}

// calculate checksum

```

sds/src/secondary_validation/france_ssn_checksum.rs
```diff
@@ -0,0 +1,92 @@
...
+                }
+            }
+        }
+        if count != FRANCE_SSN_LENGTH {
+            return false;
+        }
+
+        let final_value = first_value - penalties;
+
+        (97 - final_value % 97) == control_key
```
fuchsnj, line 53: `97` should either be made a constant, or you can also just directly use the ascii value of `a` with `b'a'` (and casting to integer type you need)

sds/src/secondary_validation/france_ssn_checksum.rs
isabella-garza-datadog: Agreed, I also thought it was a little convoluted. I wasn't sure of a better way to do it in Rust so thanks for the example!

sds/src/secondary_validation/france_ssn_checksum.rs
isabella-garza-datadog: Gotcha, I will make it a constant. Though I am a little confused what you mean by

> you can also just directly use the ascii value of a with b'a' (and casting to the integer type you need)

What would `a` be in this case?

fuchsnj: I'm just assuming `97` is being used since it's the ASCII value of `a` (which is pretty typical for checksums like this), so you could just directly use `b'a'` (which is 97) instead of making a separate constant.

fuchsnj (approved):
This is much easier to read, thanks for the refactoring.

PR #179, opened 2025-05-20
Add checksum for dutch passport

Taiki-San:


sds/src/secondary_validation/dutch_passport_checksum.rs
```diff
@@ -0,0 +1,101 @@
...
+         * - First two characters are letters from A-N or P-Z (excluding O)
+         * - Next 6 characters can be either digits or letters from A-N or P-Z (excluding O)
+         * - Last character is a checksum digit calculated using the ICAO 9303 algorithm
+         */
+        if regex_match.len() != DUTCH_PASSPORT_LENGTH {
+            return false;
+        }
+
+        // Check first two characters are valid letters (A-N or P-Z)
+        if !regex_match[..2].chars().all(|c| {
```
fuchsnj, line 20: Direct indexing of a string is a bit dangerous unless you have checked that the string contains only ASCII chars, since some Unicode characters can be multiple bytes and if you index in the middle of a character you will get a panic in Rust. The regex this checksum is used with _may_ be enforcing this already, but it's best to avoid this. In this case it's easy to just use the `.chars()` iterator, in addition to `.take()` to take a certain number of chars.

sds/src/secondary_validation/dutch_passport_checksum.rs
```diff
@@ -0,0 +1,101 @@
...
+            } else {
+                let upper = c.to_ascii_uppercase();
+                upper.is_ascii_alphabetic() && upper != 'O'
+            }
+        }) {
+            return false;
+        }
+
+        // Last character must be a digit (checksum)
+        if !regex_match.chars().last().unwrap().is_ascii_digit() {
```
fuchsnj, line 40: Similarly, only the `.len()` of the string was checked, which is the number of _bytes_, but this `.unwrap()` is asserting that the number of _characters_ which will panic if there is a multi-byte char anywhere in the match.

sds/src/secondary_validation/dutch_passport_checksum.rs
```diff
@@ -0,0 +1,101 @@
...
+            return false;
+        }
+
+        // Convert characters to their numeric values for checksum calculation
+        // A-N = 10-23, P-Z = 24-35
+        let mut numeric_values = Vec::with_capacity(8);
```
fuchsnj, line 46: Instead of allocating a `Vec` and storing all of the `numeric_values`, they can be immediately summed instead.

fuchsnj (commented):
I'll push a commit with the fixes for my suggestions (since this is from the rule rush), just wanted to give the full review so you can get feedback.

PR #180, opened 2025-05-20
Add Irish PPS checksum

fuchsnj:


sds/src/scanner/regex_rule/config.rs
```diff
@@ -148,6 +148,7 @@ pub enum SecondaryValidator {
     PolishNationalIdChecksum,
     LuxembourgIndividualNINChecksum,
     FranceSsnChecksum,
+    IrisPPS,
```
m-shykula, line 151: `IrishPPS` 

`-5` points for Gryffindor!

vinckama (commented):
I added a few suggestions. I will add them in a single commit

PR #181, opened 2025-05-20
[SDS] Add Italian Fiscale Code checksum

artslidd:
## Description

Find the description of the checksum here: https://en.wikipedia.org/wiki/Italian_fiscal_code (Section Check Character)

### DIFF
--- a/sds/src/secondary_validation/italian_national_id_checksum.rs
+++ b/sds/src/secondary_validation/italian_national_id_checksum.rs
@@ -1,0 +1,117 @@
+use crate::secondary_validation::Validator;
+use std::collections::HashMap;
+
+pub struct ItalianNationalIdChecksum;
+
+const ITALIAN_NATIONAL_ID_LENGTH: usize = 16;
+const ALLOWED_CHARACTERS: &str = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ";
+
+use lazy_static::lazy_static;
+
+lazy_static! {
+    static ref ODD_CHARACTERS_MAPPING: HashMap<char, u32> = HashMap::from([
+        ('0', 1),
+        ('1', 0),
+        ('2', 5),
+        ('3', 7),
+        ('4', 9),
+        ('5', 13),
+        ('6', 15),
+        ('7', 17),
+        ('8', 19),
+        ('9', 21),
+        ('A', 1),
+        ('B', 0),
+        ('C', 5),
+        ('D', 7),
+        ('E', 9),
+        ('F', 13),
+        ('G', 15),
+        ('H', 17),
+        ('I', 19),
+        ('J', 21),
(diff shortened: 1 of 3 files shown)


sds/src/secondary_validation/italian_national_id_checksum.rs
```diff
@@ -0,0 +1,117 @@
...
+impl Validator for ItalianNationalIdChecksum {
+    fn is_valid_match(&self, regex_match: &str) -> bool {
+        if regex_match.len() != ITALIAN_NATIONAL_ID_LENGTH {
+            return false;
+        }
+
+        // Check if the string contains only allowed characters
+        if !regex_match.chars().all(|c| ALLOWED_CHARACTERS.contains(c)) {
+            return false;
+        }
```
vinckama, line 61: If we even add a optional separators in the regex, then the checksum will fail. You could assume that all characters are valid + the length is ok (it's the role of the regex). Any non valid character can be filtered out with `        let valid_chars = regex_match.chars().filter(|c| c.is_ascii_alphanumeric());`

sds/src/secondary_validation/italian_national_id_checksum.rs
```diff
@@ -0,0 +1,117 @@
...
+            }
+
+            if position % 2 == 1 {
+                // In case of odd position, we need to use the mapping to get the checksum value
+                checksum_value += ODD_CHARACTERS_MAPPING[&c];
+            } else {
+                // In case of even position, we need to handle the character differently
+                if c.is_ascii_digit() {
+                    // If the character is a digit, add it to the checksum value
+                    checksum_value += c.to_digit(10).unwrap();
```
vinckama, line 81: Using Some in if close is a bit more idiomatic. The comment can also be removed because it describes the code line below (instead of explaining why it is needed, code without should be sufficient in most case)
```suggestion
                if let Some(digit) = c.to_digit(10) {
                    checksum_value += digit
                }
```

sds/src/secondary_validation/italian_national_id_checksum.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        let mut checksum_char = '0';
+        let mut checksum_value = 0;
+
+        for (idx, c) in regex_match.chars().enumerate() {
+            let position = idx + 1;
+            if position == ITALIAN_NATIONAL_ID_LENGTH {
+                // Skip the last character which is the checksum character
+                checksum_char = c;
+                continue;
```
vinckama, line 71: If it's the last character, then the loop can be break, this is particularly useful when/if `regex_match.len() != ITALIAN_NATIONAL_ID_LENGTH` is removed
```suggestion
                break;
```

vinckama (commented):
I'm going to apply the comments locally. 👍

PR #184, opened 2025-05-20
Romania personal numeric code

fuchsnj:


sds/src/secondary_validation/romanian_personal_numeric_code.rs
```diff
@@ -0,0 +1,79 @@
...
+impl Validator for RomanianPersonalNumericCode {
+    fn is_valid_match(&self, regex_match: &str) -> bool {
+        if !regex_match.is_ascii() {
+            return false;
+        }
+        for c in regex_match.chars() {
+            if !c.is_ascii_digit() {
+                return false;
+            }
+        }
```
vinckama, line 15: We could make this checksum a bit more future proof by removing the separators instead of declaring the checksum as non valid if there is any separator (the regex should check the number of digits in any case.)

This way, filter_map could he used here
```rust
```

sds/src/secondary_validation/romanian_personal_numeric_code.rs
```diff
@@ -0,0 +1,79 @@
...
+            calculated_checksum = 1;
+        }
+
+        let actual_checksum = if let Some(c) = regex_match.chars().last().unwrap().to_digit(10) {
+            c
+        } else {
+            return false;
+        };
+
+        calculated_checksum == actual_checksum
```
vinckama, line 38: We should ideally avoid unwrap. 
Assuming the digits have been extracted from the input, this could be simplified with a `next`:

```rust
        if let Some(actual_checksum) = digits.next() {
            return calculated_checksum == actual_checksum;
        } 
        false
```

sds/src/secondary_validation/romanian_personal_numeric_code.rs
```diff
@@ -0,0 +1,79 @@
...
+            return false;
+        }
+        for c in regex_match.chars() {
+            if !c.is_ascii_digit() {
+                return false;
+            }
+        }
+
+        let mut calculated_checksum = 0;
+        for (i, c) in regex_match.chars().take(12).enumerate() {
```
vinckama, line 18: We could use `CHECKSUM_WEIGHTS.length()` instead of `12` here to avoid any issue between the two constants

vinckama (commented):
I added a few suggestions, I will apply them in a single commit.

PR #187, opened 2025-05-20
Lithuanian personal identification number checksum

fuchsnj:


sds/src/secondary_validation/lithuanian_personal_identification_number_checksum.rs
```diff
@@ -0,0 +1,72 @@
...
+const ROUND_1_WEIGHTS: &[u32] = &[1, 2, 3, 4, 5, 6, 7, 8, 9, 1];
+const ROUND_2_WEIGHTS: &[u32] = &[3, 4, 5, 6, 7, 8, 9, 1, 2, 3];
+
+pub struct LithuanianPersonalIdentificationNumberChecksum;
+
+impl Validator for LithuanianPersonalIdentificationNumberChecksum {
+    fn is_valid_match(&self, regex_match: &str) -> bool {
+        if regex_match.chars().any(|c| !c.is_ascii_digit()) {
+            return false;
+        }
```
vinckama, line 12: We could ignore separators to make this a bit more future proof in case the regex is updated in the future to contain some of them.

As all characters are digits, they could be converted right away with 
```rust
```

sds/src/secondary_validation/lithuanian_personal_identification_number_checksum.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        let mut checksum = run_round(regex_match, ROUND_1_WEIGHTS) % 11;
+        if checksum == 10 {
+            checksum = run_round(regex_match, ROUND_2_WEIGHTS) % 11;
+            if checksum == 10 {
+                checksum = 0;
+            }
+        }
+
+        let actual_checksum = regex_match.chars().last().unwrap().to_digit(10).unwrap();
```
vinckama, line 27: Instead of validating the length above and then using unwrap here, we could take the first 10 digits above and then use `digits.next()` to validate that the input is composed of 11 digits:

```rust
        let actual_checksum = match digits.next() {
            Some(checksum) => checksum,
            None => return false,
        };
```

sds/src/secondary_validation/lithuanian_personal_identification_number_checksum.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+fn run_round(value: &str, weights: &[u32]) -> u32 {
+    let mut checksum = 0;
+
+    for (i, c) in value.chars().take(10).enumerate() {
+        checksum += weights[i] * c.to_digit(10).unwrap();
+    }
+
+    checksum
```
vinckama, line 40: If the input is converted to digits right away, then a slice to be used as the value as it should simplify this code:

```suggestion
fn run_round(value: &[u32], weights: &[u32; 10]) -> u32 {
    let mut checksum = 0;
    for (digit, weight) in value.iter().zip(weights) {
        checksum += digit * weight;
    }
    checksum
```

vinckama (commented):

PR #189, opened 2025-05-20
Add the Finnish Hetu checksum computation

Taiki-San:


sds/src/secondary_validation/finnish_hetu_checksum.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        // Validate individual number
+        if !individual_number.chars().all(|c| c.is_ascii_digit()) {
+            return false;
+        }
+
+        // Validate control character
+        if !CONTROL_CHARS.contains(control_char.to_ascii_uppercase()) {
+            return false;
+        }
```
vinckama, line 53: The regex should already check these characters, the checksum can only focus on validating the control character.

PR #191, opened 2025-05-20
Swedish PIN checksum

fbryden:
- [JIRA link](https://datadoghq.atlassian.net/browse/SDS-743)

### DIFF
--- a/sds/src/secondary_validation/sweden_pin_checksum.rs
+++ b/sds/src/secondary_validation/sweden_pin_checksum.rs
@@ -1,0 +1,72 @@
+pub struct SwedenPINChecksum;
+use crate::secondary_validation::Validator;
+
+const MULTIPLIERS: [u32; 9] = [2, 1, 2, 1, 2, 1, 2, 1, 2];
+
+fn sum_of_digits(digits: u32) -> u32 {
+    let mut sum = 0;
+    let mut num = digits;
+    while num > 0 {
+        sum += num % 10;
+        num /= 10;
+    }
+    sum
+}
+
+impl Validator for SwedenPINChecksum {
+    fn is_valid_match(&self, regex_match: &str) -> bool {
+        /*
+         * Swedish Persional Identification Number uses a 13-digit identification code.
+         * The first 9 digits are multiplied by a multiplier from the MULTIPLIERS array.
+         * Each product has its digits summed up and and added to a total.
+         * The total is used to compute a checksum, which is compared to the last digit of the input.
+         */
+
+        let mut total = 0;
+        let mut index = 0;
+        let mut checksum_digit = 0;
+
+        for char in regex_match.chars() {
+            if let Some(digit) = char.to_digit(10) {
+                if index == MULTIPLIERS.len() {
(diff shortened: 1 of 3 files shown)


sds/src/secondary_validation/sweden_pin_checksum.rs
```diff
@@ -0,0 +1,72 @@
...
+         * Swedish Persional Identification Number uses a 13-digit identification code.
+         * The first 9 digits are multiplied by a multiplier from the MULTIPLIERS array.
+         * Each product has its digits summed up and and added to a total.
+         * The total is used to compute a checksum, which is compared to the last digit of the input.
+         */
+
+        let mut total = 0;
+        let mut index = 0;
+        let mut checksum_digit = 0;
+
```
vinckama, line 28: If we even add a optional separators in the regex, then the checksum will fail. To make the checksum a bit more flexible, any non valid character can be filtered out with         `let valid_chars = regex_match.chars().filter(|c| c.is_ascii_digit())`;

sds/src/secondary_validation/sweden_pin_checksum.rs
```diff
@@ -0,0 +1,72 @@
...
+                    checksum_digit = digit;
+                    break;
+                }
+                let product = digit * MULTIPLIERS[index];
+                println!(
+                    "Currently on digit: {} product: {}, sum of digits: {}",
+                    digit,
+                    product,
+                    sum_of_digits(product)
+                );
```
vinckama, line 42: Nit:
```suggestion
```

sds/src/secondary_validation/sweden_pin_checksum.rs
```diff
@@ -0,0 +1,72 @@
...
+
+fn sum_of_digits(digits: u32) -> u32 {
+    let mut sum = 0;
+    let mut num = digits;
+    while num > 0 {
+        sum += num % 10;
+        num /= 10;
+    }
+    sum
+}
```
vinckama, line 14: This will be useful for a few other checksums, I will move it to `mod.rs` where we already have a few other helper methods 👍

vinckama (commented):
I'm going to apply the comments locally. 👍

sds/src/secondary_validation/sweden_pin_checksum.rs
```diff
@@ -0,0 +1,72 @@
...
+         * The first 9 digits are multiplied by a multiplier from the MULTIPLIERS array.
+         * Each product has its digits summed up and and added to a total.
+         * The total is used to compute a checksum, which is compared to the last digit of the input.
+         */
+
+        let mut total = 0;
+        let mut index = 0;
+        let mut checksum_digit = 0;
+
+        for char in regex_match.chars() {
```
vinckama: You could use `enumerate` to avoid having to manage the index

sds/src/secondary_validation/sweden_pin_checksum.rs
vinckama: I decided to use `filter_map` to simplify it even more

PR #192, opened 2025-05-20
[SDS-summit] Latvian National Identification Number checksum

OBe95:
Latvia national identification number has 2 formats:

1. Since July 1st 2017, an 11-digit number, starting with "32" with possible '-' after the 6th digit
2. Before July 1st 2017, an 11 digits number, in the format DDMMYY-XNNNZ, where:
  a. `DDMMYY` is the date of birth
  b. `X` represents the century digit (0 for 19, 1 for 20, 2 for 21)
  c. `NNN` is a birth serial number in that day
  d. `Z` is the checksum digit

So we only need to check the validity of `Z` when the match doesn't start with `32`


vinckama (commented):

PR #194, opened 2025-05-20
[SDS] Add Polish NIP checksum

artslidd:
## Description

Source: https://github.com/aashishvanand/corporate-taxid-checker-js/blob/main/src/validators/PL.js#L2

Implements the Polish NIP checksum

### DIFF
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -686,7 +686,6 @@ impl ScannerBuilder<'_> {
     }
 
     pub fn build(self) -> Result<Scanner, CreateScannerError> {
-        let mut scanner_features = self.scanner_features.clone();
         let mut match_validators_per_type = AHashMap::new();
 
         for rule in self.rules.iter() {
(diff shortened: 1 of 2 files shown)


sds/src/secondary_validation/polish_nip_checksum.rs
```diff
@@ -0,0 +1,56 @@
...
+impl Validator for PolishNipChecksum {
+    fn is_valid_match(&self, regex_match: &str) -> bool {
+        if regex_match.len() != POLISH_NIP_LENGTH {
+            return false;
+        }
+
+        // Check if the string contains only allowed characters
+        if !regex_match.chars().all(|c| c.is_ascii_digit()) {
+            return false;
+        }
```
vinckama, line 18: We could make this checksum a bit more future proof by removing the separators instead of declaring the checksum as non valid if there is any separator (the regex should check the number of digits in any case.)

This way, filter_map could he used here
```rust
```

vinckama (commented):
I'm going to apply similar suggestions  to https://github.com/DataDog/dd-sensitive-data-scanner/pull/184

PR #198, opened 2025-05-20
[SDS] Add Spanish DNI checksum

artslidd:
## Description

Find the algo descripted here: https://en.wikipedia.org/wiki/National_Identity_Card_(Spain)#Number

### DIFF
--- a/sds/src/secondary_validation/spain_dni_checksum.rs
+++ b/sds/src/secondary_validation/spain_dni_checksum.rs
@@ -2,12 +2,11 @@ use crate::secondary_validation::Validator;
 
 pub struct SpanishDniChecksum;
 
-const LETTER_TABLE: [&str; 23] = [
-    "T", "R", "W", "A", "G", "M", "Y", "F", "P", "D", "X", "B", "N", "J", "Z", "S", "Q", "V", "H",
-    "L", "C", "K", "E",
+const LETTER_TABLE: [char; 23] = [
+    'T', 'R', 'W', 'A', 'G', 'M', 'Y', 'F', 'P', 'D', 'X', 'B', 'N', 'J', 'Z', 'S', 'Q', 'V', 'H',
+    'L', 'C', 'K', 'E',
 ];
 
-const DNI_LENGTH: usize = 9;
 const NUMBER_LENGTH: usize = 8;
 impl Validator for SpanishDniChecksum {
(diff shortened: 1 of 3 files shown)


sds/src/secondary_validation/spain_dni_checksum.rs
```diff
@@ -0,0 +1,75 @@
...
+         * 8 digits followed by a letter
+         * The letter is calculated by taking the number modulo 23 and using it as an index
+         * into LETTER_TABLE
+         */
+        if regex_match.len() != DNI_LENGTH {
+            return false;
+        }
+
+        // Get the number and letter parts
+        let (number_part, letter_part) = regex_match.split_at(8);
```
vinckama, line 25: We could panic if the input is not only ascii characters. This can be replaced by 
`input.by_ref().take(NUMBER_LENGTH).collect::<String>()` to take the first 8 chars.

sds/src/secondary_validation/spain_dni_checksum.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        // Check if first 8 chars are digits and last is a letter
+        if !number_part.chars().all(|c| c.is_ascii_digit()) {
+            return false;
+        }
```
vinckama, line 30: This check should be done by the regex, we could avoid it here. We could also filter out all separators to make the match validation more flexible (in case the regex contains separators)

sds/src/secondary_validation/spain_dni_checksum.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        let number_part: u32 = regex_match[..NUMBER_LENGTH].parse().unwrap();
+
+        // Calculate expected letter
+        let index = number_part % 23;
+        let expected_letter = LETTER_TABLE[index as usize];
```
vinckama, line 36: To avoid any issue:
```suggestion
        let index = number_part % LETTER_TABLE.len();
        let expected_letter = LETTER_TABLE[index as usize];
```

sds/src/secondary_validation/spain_dni_checksum.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        letter_part == expected_letter
```
vinckama, line 38: I think the matching should be case-insensitive in case the regex is.

vinckama (commented):

PR #206, opened 2025-06-05
[SDS-1100] Always explicitly require with_return_matches(true)

zachchentouf:
[OG Comment
](https://github.com/DataDog/dd-sensitive-data-scanner/pull/177#discussion_r2067269068)

Now, we:
- Explicitly require with_return_matches(true) to enable the feature while creating the scanner. This way, with_return_matches will be visible on the user code and validate_matches will always fail if return_matches is false.
- Update tests to reflect this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,13 @@ target/
 tools/fuzz/out/
 **/.DS_Store
 
-.cursor/
+# Cursor Profiles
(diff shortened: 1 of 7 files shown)


fbryden (approved):
Very nice, I feel like this solves the problem statement correctly ✅

PR #207, opened 2025-06-05
Temporary workaround for the regex cache unwrap panic

fuchsnj:
Somehow the `.unwrap()` of `raw_get` can panic, and the root cause hasn't been identified yet. As a temporary workaround, in the case the cache is not available for a regex, a temporary one is created. The main downside here is creating a new cache is way more expensive than re-using one, but it does avoid a panic. As long as this doesn't happen very often the overall perf impact should be negligible.

Since the panic will no longer be visible, a new metric was created to track this error.


fuchsnj:
The root cause here was memory corruption due to unrelated code. This is no longer needed.

PR #227, opened 2025-07-03
use a local rayon thread pool

fuchsnj:
By default `rayon` will use a global thread pool with configuration of whatever code configures it first, potentially causing conflicts (e.g. different number of workers threads than expected). To prevent issues, a local thread pool is created and used.

### DIFF
--- a/sds/src/match_validation/http_validator.rs
+++ b/sds/src/match_validation/http_validator.rs
@@ -2,11 +2,12 @@ use super::{
     config::{CustomHttpConfig, RequestHeader},
     match_validator::MatchValidator,
 };
+use crate::match_validation::match_validator::RAYON_THREAD_POOL;
 use crate::{match_validation::config::HttpMethod, MatchStatus, RuleMatch};
 use crate::{scanner::RootCompiledRule, HttpValidatorOption};
 use ahash::AHashMap;
 use lazy_static::lazy_static;
-use rayon::prelude::*;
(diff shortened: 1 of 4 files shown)


sds/src/match_validation/http_validator.rs
```diff
@@ -2,11 +2,11 @@ use super::{
     config::{CustomHttpConfig, RequestHeader},
     match_validator::MatchValidator,
 };
+use crate::match_validation::match_validator::RAYON_THREAD_POOL;
 use crate::{match_validation::config::HttpMethod, MatchStatus, RuleMatch};
 use crate::{scanner::RootCompiledRule, HttpValidatorOption};
 use ahash::AHashMap;
 use lazy_static::lazy_static;
-use rayon::prelude::*;
```
vinckama, line 9: Could/Do we have a way to enforce that `use rayon::prelude::*;` is never used globally?
This can probably be tackled as follow-up PR

vinckama (commented):
LGTM with question for follow-ups

sds/src/match_validation/http_validator.rs
fuchsnj: I'm not sure of a feasible way to enforce this. The global import itself doesn't actually change anything. The `RAYON_THREAD_POOL.install` is the important part. I just moved the imports to make it slightly harder to mess it up. Maybe other concurrency libraries make this easier.

vinckama (approved):

PR #231, opened 2025-07-11
[SDS-1326] Default hosts to empty list when deserializing

artslidd:
When deserializing the third party active checker config, make sure we default the hosts attribute to an empty list. 

Add some tests to validate the behaviour.

### DIFF
--- a/sds/src/match_validation/config.rs
+++ b/sds/src/match_validation/config.rs
@@ -100,7 +100,7 @@ pub struct HttpValidatorOption {
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 pub struct CustomHttpConfig {
     pub endpoint: String,
-    #[serde(default = "default_hosts")]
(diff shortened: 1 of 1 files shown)


sds/src/match_validation/config.rs
```diff
@@ -100,7 +100,7 @@ pub struct HttpValidatorOption {
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 pub struct CustomHttpConfig {
     pub endpoint: String,
-    #[serde(default)]
+    #[serde(default = "default_hosts")]
```
fuchsnj, line 103: You can just use `#[serde(default)]` if you want since `Vec` [implements](https://doc.rust-lang.org/src/alloc/vec/mod.rs.html#3811) `Default` (which is an empty Vec)

fuchsnj (approved):

PR #232, opened 2025-07-16
cleanup get_string_matches function

fuchsnj:
This replaces 8 of the `get_string_matches` arguments with a struct to clean up things a bit in preparation for some new `async` versions of these functions.


zachchentouf (approved):
Looks straightforward to me. Just curious why it was setup like this in the first place / impetus for the cleanup?

PR #234, opened 2025-07-23
Add validation function to sds-go

fbryden:
* Add caching to our validation function in our dd-sds engine
* Add a binding for that same validation function in Go


sds/src/validation.rs
```diff
@@ -41,7 +46,14 @@ const REGEX_COMPLEXITY_LIMIT: usize = 1_000_000;
 pub fn validate_regex(input: &str) -> Result<(), RegexValidationError> {
     // This is the same as `validate_and_create_regex`, but removes the actual Regex type
     // to create a more stable API for external users of the crate.
-    validate_and_create_regex(input).map(|_| ())
+    if let Some(cache_hit) = REGEX_CACHE.get(input) {
```
fbryden, line 49: The get is run separately (instead of running a single `get_with` call which combines both since:
* Get does not require an owned type and can compare with the reference
* This way we avoid cloning on each iteration, especially since the cache hit rate will quickly become very high

sds-go/go/validation_test.go
```diff
@@ -0,0 +1,22 @@
...
+		t.Fatal(err)
+	}
+	fmt.Println("valid:", valid)
+
+	// Test invalid regex to ensure memory is properly freed
+	valid, err = ValidateRegex("[")
+	if err == nil {
+		t.Fatal("Expected error for invalid regex")
+	}
+	fmt.Println("invalid regex error:", err)
```
artslidd, line 21:
```suggestion

```

sds-go/go/validation_test.go
```diff
@@ -0,0 +1,22 @@
...
+	"testing"
+)
+
+func TestValidateRegex(t *testing.T) {
+	// Test valid regex
+	valid, err := ValidateRegex("hello")
+	if err != nil {
+		t.Fatal(err)
+	}
+	fmt.Println("valid:", valid)
```
artslidd, line 14:
```suggestion
```

sds-go/go/validation_test.go
fbryden: Done in a072629

sds/src/validation.rs
fbryden: I'm worrying about cloning the string, `input`, not the error 🙈

fuchsnj: Ah, sorry. Didn't read it close enough. 👍

PR #237, opened 2025-07-30
Add a JWT checker

Taiki-San:
The checker can be configured to check whether claims match specific values or regexes.
That's because we need to check two different JWTs and this avoid having to introduce too many different checkers

### DIFF
--- a/sds/src/scanner/regex_rule/config.rs
+++ b/sds/src/scanner/regex_rule/config.rs
@@ -144,19 +144,11 @@ pub enum ClaimRequirement {
 }
 
 #[serde_as]
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[derive(Serialize, Deserialize, Default, Clone, Debug, PartialEq)]
 pub struct JwtClaimsCheckerConfig {
     #[serde_as(deserialize_as = "DefaultOnNull")]
     #[serde(default)]
     pub required_claims: std::collections::HashMap<String, ClaimRequirement>,
-}
-
-impl Default for JwtClaimsCheckerConfig {
-    fn default() -> Self {
-        Self {
-            required_claims: std::collections::HashMap::new(),
-        }
-    }
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, EnumIter, AsRefStr)]
--- a/sds/src/scanner/regex_rule/config.rs
+++ b/sds/src/scanner/regex_rule/config.rs
@@ -133,6 +133,32 @@ pub struct ProximityKeywordsConfig {
     pub excluded_keywords: Vec<String>,
 }
 
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+pub enum ClaimRequirement {
+    /// Just check that the claim exists
+    Present,
+    /// Check that the claim exists and has an exact value
+    ExactValue(String),
+    /// Check that the claim exists and matches a regex pattern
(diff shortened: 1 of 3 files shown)


sds/src/scanner/regex_rule/config.rs
```diff
@@ -133,6 +133,32 @@ pub struct ProximityKeywordsConfig {
...
+    pub required_claims: std::collections::HashMap<String, ClaimRequirement>,
+}
+
+impl Default for JwtClaimsCheckerConfig {
+    fn default() -> Self {
+        Self {
+            required_claims: std::collections::HashMap::new(),
+        }
+    }
+}
```
ikretz, line 160: You can `derive` this default behavior: `#derive(..., Default, ...)`

sds/src/secondary_validation/jwt_claims_checker.rs
```diff
@@ -0,0 +1,228 @@
...
+    if raw_segments.len() != SEGMENTS_COUNT {
+        // Invalid JWT header + payload + signature
+        return None;
+    }
+
+    let header_segment = raw_segments[0];
+    let payload_segment = raw_segments[1];
+    let header_json = decode_segment(header_segment)?;
+    let payload_json = decode_segment(payload_segment)?;
+    Some((header_json, payload_json))
```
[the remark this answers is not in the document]
Taiki-San: The regex check that it exists, but unless we want to keep an up to date public key to validate the signature, there isn't much to do that the regex isn't already doing

sds/src/secondary_validation/mod.rs
```diff
@@ -173,6 +175,9 @@ impl Validator for SecondaryValidator {
             SecondaryValidator::ItalianNationalIdChecksum => {
                 ItalianNationalIdChecksum.is_valid_match(regex_match)
             }
+            SecondaryValidator::JwtClaimsChecker { config } => {
+                JwtClaimsChecker::new(config.clone()).is_valid_match(regex_match)
```
fuchsnj, line 179: This is cloning the config and creating a new `JwtClaimsChecker` every time the validator runs. The main problem is right now the code assumes that the "Config" that is directly deserialized is the validator that will run since all validators until now have been lightweight / don't need any compilation. We will either need to change this a bit to allow a "compilation" step, or do a more short term fix here to cache the compilation. But right now this is still compiling the regex / other setup for every validation call which is way too expensive.

sds/src/secondary_validation/jwt_claims_checker.rs
```diff
@@ -0,0 +1,233 @@
...
+        .ok()?;
+    serde_json::from_slice(&decoded).ok()
+}
+
+fn validate_required_claims(payload: &JsonValue, required_claims: &Vec<(String, ClaimRequirement)>, patterns: &AHashMap<String, Regex>) -> bool {
+    if let Some(payload_obj) = payload.as_object() {
+        // Check each required claim
+        required_claims.iter().all(|(claim_name, requirement)| {
+            if let Some(claim_value) = payload_obj.get(claim_name) {
+                validate_claim_requirement(claim_value, requirement, patterns.get(claim_name))
```
fuchsnj, line 68: nit: It's probably a bit easier to pass in a reference to the entire `patterns` map, so `validate_claim_requirement` can check for the regex pattern only if it's needed. Right now it fetches it for every claim.

fuchsnj (approved):
One minor comment, nothing blocking though. Thanks for the improvements!

PR #238, opened 2025-07-30
Use hybrid data structure for `InternalRuleMatchSet`

fuchsnj:
This should improve the performance a bit, especially if there are no async matches. This also guarantees the ordering of sync matches. Async matches will always be at the end.

### DIFF
--- a/sds/src/scanner/internal_rule_match_set.rs
+++ b/sds/src/scanner/internal_rule_match_set.rs
@@ -3,31 +3,52 @@ use crate::{Encoding, Path};
 use ahash::AHashMap;
 
 pub struct InternalRuleMatchSet<E: Encoding> {
+    sync_matches: Vec<(Path<'static>, Vec<InternalRuleMatch<E>>)>,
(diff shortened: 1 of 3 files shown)


sds/src/scanner/internal_rule_match_set.rs
```diff
@@ -3,31 +3,52 @@ use crate::{Encoding, Path};
...
+        if !self.async_matches.is_empty() {
+            // merge async matches into sync matches if there are matches for the same path
+            for (path, matches) in &mut self.sync_matches {
+                if let Some(async_matches) = self.async_matches.remove(path) {
+                    matches.extend(async_matches)
+                }
+            }
+        }
+
+        self.sync_matches.into_iter().chain(self.async_matches)
```
trosenblatt, line 52: So potentially we won't have all the matches per path sequentially

fuchsnj: They are merged above, so you will receive all matches for each path once

trosenblatt (approved):

PR #240, opened 2025-07-31
[SDS-1310] - Allow Map Scans to Scan for booleans and numbers

zachchentouf:
Small change to allow more scanning in dd-sds for agentless
This will allow agentless scanners to scan JSON maps that contain numbers and not just strings.


zachchentouf:
Not needed for now

zachchentouf:
CC @isabella-garza-datadog

PR #246, opened 2025-08-05
add "compilation" step for validators

fuchsnj:
[Jira](https://datadoghq.atlassian.net/browse/SDS-1411)

Some future secondary validators (e.g. https://github.com/DataDog/dd-sensitive-data-scanner/pull/237) will need a "compilation step" (e.g. compiling a regex) that is done once before the validator can be used. This adds support for that by adding a `compile` function to `SecondaryValidator` that returns a `dyn Validator`.


sds/src/secondary_validation/mod.rs
```diff
@@ -123,105 +124,67 @@ fn sum_all_digits(digits: u32) -> u32 {
...
+                Arc::new(CzechTaxIdentificationNumberChecksum)
+            }
+            SecondaryValidator::DutchBsnChecksum => Arc::new(DutchBsnChecksum),
+            SecondaryValidator::DutchPassportChecksum => Arc::new(DutchPassportChecksum),
+            SecondaryValidator::EthereumChecksum => Arc::new(EthereumChecksum),
+            SecondaryValidator::FinnishHetuChecksum => Arc::new(FinnishHetuChecksum),
+            SecondaryValidator::FranceNifChecksum => Arc::new(FranceNifChecksum),
+            SecondaryValidator::FranceSsnChecksum => Arc::new(FranceSsnChecksum),
+            SecondaryValidator::GermanIdsChecksum => Arc::new(GermanIdsChecksum),
+            SecondaryValidator::GermanSvnrChecksum => Arc::new(GermanSvnrChecksum),
```
fbryden, line 150: It's a shame there's no better way of doing this 😬

fbryden (approved):
LGTM!

sds/src/secondary_validation/mod.rs
fuchsnj: There's a crate we can pull in that will make it easier. It essentially writes this as compile time for us. I'll add it to my todo list to simplify this

PR #247, opened 2025-08-07
Lock rust version

fuchsnj:
This locks the `rustc` version to 1.89. We will still keep updating the version, but this allows us to choose when it's updated and fix clippy / formatting errors at the same time, rather than whenever the releases comes out, which may make arbitrary PRs fail in CI.

### DIFF
--- a/sds/src/parser/regex_parser.rs
+++ b/sds/src/parser/regex_parser.rs
@@ -209,7 +209,7 @@ fn unicode_property_class(input: Input) -> ParseResult<UnicodePropertyClass> {
     ))(input)
 }
 
-fn unicode_property_name(input: Input) -> ParseResult<&str> {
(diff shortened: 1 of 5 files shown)


sds/src/parser/regex_parser.rs
```diff
@@ -209,7 +209,7 @@ fn unicode_property_class(input: Input) -> ParseResult<UnicodePropertyClass> {
     ))(input)
 }
 
-fn unicode_property_name(input: Input) -> ParseResult<&str> {
+fn unicode_property_name(input: Input<'_>) -> ParseResult<'_, &str> {
```
isabella-garza-datadog, line 212: Just curious, what is this change doing and why don't we need it for all of the functions in this file that also take an `Input` type and output `ParseResult`?

isabella-garza-datadog (approved):
changes lgtm! just left one question

PR #250, opened 2025-08-11
[SDS-1412] Implement Optional Match Status / 3rd Party Active Checker

zachchentouf:
# Use `with_validate_matching`(with default to false) in Go bindings

## Summary
This PR updates the Go bindings to make the `with_validate_matching` parameter default to `false` for backward compatibility and improved performance in the default case.

Will then need to be used in `datadog-agentless-scanner`.

### Go Bindings API Changes
- **Modified `Scan` and `ScanEventsMap` methods** to use `with_validate_matching=false` by default
  - `Scan(event string)` now calls `ScanWithValidation(event, false)` internally
  - `ScanEventsMap(event map[string]interface{})` now calls `ScanEventsMapWithValidation(event, false)` internally
  
- **Added explicit validation methods** for users who need match validation:
  - `ScanWithValidation(event string, with_validate_matching bool)` 
  - `ScanEventsMapWithValidation(event map[string]interface{}, with_validate_matching bool)`

### Binary Protocol Updates
- Updated the binary encoding/decoding to handle the new `MatchStatus` field in match results
- Modified Go decoder to maintain backward compatibility by returning empty `MatchStatus` when the value is `"NotAvailable"`
- Added support for `MatchValidationError` in error handling (error code 1)

### FFI Layer Changes
- Updated C header (`dd_sds.h`) to include `int with_validate_matching` parameter in the `scan` function
- Modified Rust FFI implementation to accept and process the validation flag using `ScanOptionBuilder`

### Core Scanner Updates  
- Refactored `Scanner::scan` method to have two variants:
  - `scan(event)` - uses default options with `with_validate_matching=false`
  - `scan_with_options(event, options)` - accepts explicit `ScanOptions`
- Fixed validation flow to operate on `RuleMatch` objects instead of `InternalRuleMatch`
- Added `MatchValidation` error variant to `ScannerError` enum

## Backward Compatibility
- **Existing code continues to work without changes** - calls to `Scan()` and `ScanEventsMap()` maintain their original behavior
- **Test compatibility preserved** - `MatchStatus` field remains empty in existing tests when validation is not performed
- **No breaking changes** to the public API

## Performance Impact
- Default behavior (validation disabled) maintains optimal performance
- Match validation only runs when explicitly requested via the new `WithValidation` methods

## Testing
- ✅ All existing Go tests pass without modification
- ✅ Go example runs successfully with the new default behavior
- ✅ Rust benchmarks compile and run correctly

## Migration Guide
For users who need match validation:
```go
// Before (if using validation - which wasn't available in the old API)
result, err := scanner.Scan(event)

// After (to explicitly enable validation)
result, err := scanner.ScanWithValidation(event, true)
```

For most users, no changes are needed as the default behavior remains optimal for performance.


sds/src/scanner/error.rs
```diff
@@ -49,6 +49,8 @@ pub enum MatchValidationError {
 pub enum ScannerError {
     #[error("Transient error while scanning")]
     Transient(String),
+    #[error("Match validation error")]
+    MatchValidation(#[from] MatchValidationError),
```
fuchsnj, line 53: Why is this new error needed? We already have individual statuses for each match for match validation (including errors)

sds/src/scanner/mod.rs
```diff
@@ -399,10 +409,23 @@ impl Scanner {
     // This function scans the given event with the rules configured in the scanner.
     // The event parameter is a mutable reference to the event that should be scanned (implemented the Event trait).
     // The return value is a list of RuleMatch objects, which contain information about the matches that were found.
+    // This version uses default scan options (no validation, no blocked rules, no wildcarded indices).
     pub fn scan<E: Event>(&self, event: &mut E) -> Result<Vec<RuleMatch>, ScannerError> {
```
zachchentouf, line 413: Same here as well

sds-go/go/scanner.go
```diff
@@ -178,15 +184,22 @@ func (s *Scanner) scanEncodedStringEvent(encodedEvent []byte) (ScanResult, error
 }
 
 // Scan sends the string event to the SDS shared library for processing.
+// withValidateMatching defaults to false.
 func (s *Scanner) Scan(event []byte) (ScanResult, error) {
```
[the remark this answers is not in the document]
fuchsnj: Yes, this seems like a reasonable approach. We probably want some kind of `ScanWithOptions` so there is a single scanner that can be configured with any possible option if we start adding too many things here, but this should be fine for now.

sds/src/scanner/error.rs
zachchentouf: Ok, I think I understand now!

zachchentouf: [Fix logic](https://github.com/DataDog/dd-sensitive-data-scanner/pull/250/commits/be7ca79883c547dfc85cb2baef8f3b7f61e22d8e)

Made this change, let me know what you think

sds/src/scanner/mod.rs
```diff
@@ -414,7 +437,7 @@ impl Scanner {
             .await
     }
 
-    pub fn scan_with_options<E: Event>(
+    pub fn scan_with_options_impl<E: Event>(
```
fuchsnj, line 440: Why is this extra `scan_with_options_impl` needed? It's the exact same signature as `scan_with_options`. It seems like it can be removed.

sds/src/scanner/mod.rs
```diff
@@ -335,27 +335,31 @@ pub struct ScanOptions {
     pub blocked_rules_idx: Vec<usize>,
     // The wildcarded_indices parameter is a map containing a list of tuples of (start, end) indices that should be treated as wildcards (for the message key only) per path.
     pub wildcarded_indices: AHashMap<Path<'static>, Vec<(usize, usize)>>,
+    pub with_validate_matching: bool,
```
fuchsnj, line 338: I think just `validate_matches` would be more clear here (to match the function it's calling). The `with` prefix is usually used for the builder method itself, not the value.
Also please add a comment to explain what this config options does.

sds/src/scanner/mod.rs
zachchentouf: Thanks, good catch. I added two extra function instead of just one when adding the call with default `ScanOptions::default()` from `scan`.

[Clean up signatures and remove scan_with_options_impl](https://github.com/DataDog/dd-sensitive-data-scanner/pull/250/commits/a9ddeb09e0c84f59c5a41640e13893680e819b44)

sds/src/scanner/mod.rs
zachchentouf: Sounds good! [Replace with_validate_matching with validate_matches instead](https://github.com/DataDog/dd-sensitive-data-scanner/pull/250/commits/c2ce5bc60832d0f92da8ef2518f22768c77134bf)

Renamed, added a comment, and also did some cleanup - let me know if you think it makes sense enough.

fuchsnj (approved):

PR #257, opened 2025-09-01
Update JWT Claim Checker to evaluate on the JWT header

Taiki-San:
In order to support specific JWTs, we need to check for the presence of an attribute in the JWT header.
This PR extends the current syntax of the JwtClaimsChecker to be able to explicitly address the header. This was done this way in order to avoid updating the syntax of the file format

### DIFF
--- a/sds/src/secondary_validation/jwt_claims_validator.rs
+++ b/sds/src/secondary_validation/jwt_claims_validator.rs
@@ -23,7 +23,8 @@ pub struct JwtClaimsValidatorConfig {
 }
 
 pub struct JwtClaimsValidator {
-    pub required_claims: Vec<(String, ClaimRequirement)>,
+    pub header_required_claims: Vec<(String, ClaimRequirement)>,
+    pub payload_required_claims: Vec<(String, ClaimRequirement)>,
     patterns: AHashMap<String, Regex>,
 }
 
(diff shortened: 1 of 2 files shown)


sds/src/secondary_validation/jwt_claims_validator.rs
```diff
@@ -35,17 +36,34 @@ impl JwtClaimsValidator {
...
+                .required_claims
+                .clone()
+                .into_iter()
+                .filter(|(claim_name, _)| claim_name.starts_with("header."))
+                .collect(),
+            payload_required_claims: config
+                .required_claims
+                .into_iter()
+                .filter(|(claim_name, _)| !claim_name.starts_with("header."))
+                .collect(),
```
vinckama, line 51: This will contribute to create a lot of magic around the format of headers. The configuration is not already used so I think it's better to update the config format and keep the logic here simpler than adding this blurry logic which will required additional context to use the validator and make the learning curve to use this validator steeper.

vinckama (commented):
I added one general comment on the format. I think it's better to update the configuration than adding this extra logic to distinguish claims for headers from claims from payload.

sds/src/secondary_validation/jwt_claims_validator.rs
```diff
@@ -18,34 +18,57 @@ pub enum ClaimRequirement {
 
 #[derive(Serialize, Deserialize, Default, Clone, Debug, PartialEq)]
 pub struct JwtClaimsValidatorConfig {
+    #[serde(default)]
+    pub required_headers: std::collections::BTreeMap<String, ClaimRequirement>,
     #[serde(default)]
     pub required_claims: std::collections::BTreeMap<String, ClaimRequirement>,
```
vinckama, line 24: As it's not yet used and this PR adds a new field, I suggest to make the other attribute even more explicit by renaming it `required_payload_claims`:

```suggestion
    pub required_payload_claims: std::collections::BTreeMap<String, ClaimRequirement>,
```

vinckama (approved):
LGTM with a (non-blocking) suggestion on naming

sds/src/secondary_validation/jwt_claims_validator.rs
Taiki-San: I like the parity between `required_headers` and `required_claims`. I thought it kept things short and unambiguous (at least when you see both next to one another). Also, payload is a very generic term so I'm not sure it'd make explicit that there is a `required_headers` if it's not present...

PR #339, opened 2026-03-30
Support indexing on arrays in body matching

fbryden:
Indexing is currently not supported in the body matching path of config v2.

This adds support for it.

Use the JSON path access syntax, so `a[0]` will access the first element contained in list `a`.

- [JIRA link](https://datadoghq.atlassian.net/browse/SDS-2441)


aimenebelfodil (approved):
LGTM thanks
- Could you add a test where root is an array?
- Another test where we do have nested array
- And update the PR description too 🙇 with the latest updates

Thanks @fbryden 🥇

PR #349, opened 2026-04-13
Introduce a AtlassianTokenChecksum validator

Taiki-San:
Introduce a new static validator to check Atlassian tokens


fbryden (approved):
LGTM

PR #352, opened 2026-04-24
Introduce a templating system to base64 encode variables

Taiki-San:
This PR introduce support for a syntax enabling base64 encoding parts of a request. This is helpful to support Basic Auth which involved base64 encoding the username and password to authenticate the request.
The syntax look like `%base64(arbitrary_string)`. Variables can be used in the transformation (`%base64($MATCH)`) but can't contain transformation instruction (`$MATCH` => `%base64(bla)` will be escaped and rendered literally).
The engine ignore unknown instructions (for instance, URL encoded strings)


isabella-garza-datadog (commented):
Instead of denoting the encoding in the key name, I noticed there's already a comment in `config_v2.rs` ([link](https://github.com/DataDog/dd-sensitive-data-scanner/blob/5ccd2861b8daca861bc8409999b46710ff9bd910/sds/src/match_validation/config_v2.rs#L76)) that suggests using an inline function syntax in the value instead:

`/// Example: {"Authorization": "Basic %base64($CLIENT_ID:$MATCH)"}`

I think that reads a bit more naturally since the transformation is right where the encoding happens and we don't need to strip the suffix from the key name. What do you think about using that approach instead?

Taiki-San:
The syntax that's documented here would definitely be better!
Back when we implemented the initial RFC, we assumed that this would be complex to implement. I didn't want to relitigate this again, and just unlock Basic Auth. If it's cool with you, I can send you a proposal of the more generic variant

sds/src/match_validation/config_v2.rs
```diff
@@ -395,6 +389,135 @@ impl TemplatedMatchString {
...
+            }
+            _ => {}
+        }
+    }
+
+    None
+}
+
+const KNOWN_TRANSFORMS: &[&str] = &["base64"];
+const KNOWN_TRANSFORM_PREFIXES: &[&str] = &["%base64("];
```
isabella-garza-datadog: nit: I am assuming that in the future we may want to expand `KNOWN_TRANSFORMS` to include additional transforms, and it could get tedious to keep `KNOWN_TRANSFORM_PREFIXES` in sync. Could `may_contain_transform` just check for the presence of '%' instead? The only cost is occasionally running `parse_template` on a string that has a % but no transform. I am not sure how frequently that happens but, if it is not often, then it would simplify the check a bit.

sds/src/match_validation/config_v2.rs
isabella-garza-datadog: Could we remove `with_template_variable` and instead only rely on `render_with_variables`? There is only one call site left of it in `http_validation_v2.rs` ([link](https://github.com/DataDog/dd-sensitive-data-scanner/blob/48e9c357a1874b79d88080cf19690109d9853d81/sds/src/match_validation/http_validator_v2.rs#L334)) and, if we migrate that, we would not have to worry about accidentally silently skipping transforms in `with_template_variable`.

sds/src/match_validation/config_v2.rs
```diff
@@ -395,6 +389,135 @@ impl TemplatedMatchString {
...
+
+fn may_contain_transform(input: &str) -> bool {
+    KNOWN_TRANSFORM_PREFIXES.iter().any(|p| input.contains(p))
+}
+
+fn apply_transform(name: &str, value: &str) -> String {
+    match name {
+        "base64" => base64::engine::general_purpose::STANDARD.encode(value),
+        _ => unreachable!("parser only accepts known transforms"),
```
isabella-garza-datadog: Could we avoid the runtime panic here by using an enum for transforms instead?
Right now if `KNOWN_TRANSFORMS` and `apply_transform` get out of sync, this would panic. With an enum, the compiler should catch any missing cases for us.

isabella-garza-datadog (commented):
Thanks for making the changes and the exhaustive tests! Just left a few small suggestions

sds/src/match_validation/config_v2.rs
Taiki-San: I experimented with this first, but I was concerned that URL encoded payloads would cause unnecessary work and memory allocation. I don't think we will introduce too many transformers and we have a test making sure those two arrays stay in sync so I figured this was more efficient.

sds/src/match_validation/config_v2.rs
```diff
@@ -377,24 +379,158 @@ pub struct PairedValidatorConfig {
...
+                    pos = content_end + 1;
+                } else {
+                    segments.push(TemplateSegment::Literal(&input[pos..pct + 1]));
+                    pos = pct + 1;
+                }
+            }
+            None => {
+                segments.push(TemplateSegment::Literal(&input[pos..]));
+                break;
+            }
```
isabella-garza-datadog, line 487: nit: Might be worth adding a couple lightweight comments here just to explain the flow a bit

isabella-garza-datadog (approved):
LGTM overall! Left one last small comment 😸

PR #354, opened 2026-04-29
[SDSP-324] Add is_supporting_rule to RootRuleConfig and RootCompiledRule

isabella-garza-datadog:
Supporting rules exist solely to provide template variables to `CustomHttpV2` match validators of other rules via `match_pairing`. Their matches are retained during `validate_matches` so HTTP pairing can reference their values, then stripped from the final output.

### DIFF
--- a/sds-go/go/scanner.go
+++ b/sds-go/go/scanner.go
@@ -18,10 +18,11 @@ import (
 import "C"
 
 var (
-	ErrUnknown            error = fmt.Errorf("unknown error")
-	ErrInvalidRegex       error = fmt.Errorf("invalid regex")
-	ErrInvalidKeywords    error = fmt.Errorf("invalid keywords")
-	ErrInvalidMatchAction error = fmt.Errorf("invalid match action")
+	ErrUnknown                      error = fmt.Errorf("unknown error")
+	ErrInvalidRegex                 error = fmt.Errorf("invalid regex")
+	ErrInvalidKeywords              error = fmt.Errorf("invalid keywords")
+	ErrInvalidMatchAction           error = fmt.Errorf("invalid match action")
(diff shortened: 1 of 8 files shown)


sds-go/go/scanner.go
```diff
@@ -18,10 +18,11 @@ import (
...
 var (
-	ErrUnknown            error = fmt.Errorf("unknown error")
-	ErrInvalidRegex       error = fmt.Errorf("invalid regex")
-	ErrInvalidKeywords    error = fmt.Errorf("invalid keywords")
-	ErrInvalidMatchAction error = fmt.Errorf("invalid match action")
+	ErrUnknown                     error = fmt.Errorf("unknown error")
+	ErrInvalidRegex                error = fmt.Errorf("invalid regex")
+	ErrInvalidKeywords             error = fmt.Errorf("invalid keywords")
+	ErrInvalidMatchAction          error = fmt.Errorf("invalid match action")
+	ErrInvalidSupportingRuleConfig error = fmt.Errorf("supporting rules cannot have a match action other than None")
```
fbryden: Nice yes this is nice!

fbryden: For the naming, would it be worth naming the error constant something more specific?
`InvalidRuleConfig` is quite generic when the text message / throwing context is only in the case of MatchAction non-None in a support rule.

fbryden (approved):
Very nice Isabella, this is exciting !!

Left a small comment, but looks good and testing seems appropriate !

sds-go/go/scanner.go
isabella-garza-datadog: Good call! I renamed it to `ErrSupportingRuleHasMatchAction`

PR #355, opened 2026-05-07
fix(scanner): move validate_matches outside block_on to prevent EnterError panic

isabella-garza-datadog:
## What

Refactors the sync scan path to avoid a panic triggered when `scan_with_options(validate=true)` is called concurrently from rayon worker threads (e.g. in the Datadog static analyzer, which scans files in parallel).

## Why

`scan_with_options` used `futures::executor::block_on(internal_scan_with_metrics(...))`. Inside that future, `validate_matches` called `RAYON_THREAD_POOL.install(|| par_iter_mut ...)`. When a rayon worker thread was already inside `block_on` (holding the `LocalPool` executor context) and the scheduler caused it to re-enter the executor context, `futures` panicked with:

```
cannot execute `LocalPool` executor from within another executor: EnterError
```

## How

Split `internal_scan` into two parts:

- **`internal_scan_collect`** (async) — collects matches only; no validation, no supporting-rule filtering.
- **`finalize_matches`** (sync) — runs `validate_matches` + drops supporting-rule matches. Must be called **outside** any futures executor.

`scan_with_options` now calls `block_on(internal_scan_collect(...))` then `finalize_matches(...)` after `block_on` returns, safely outside the LocalPool context.

`scan_async_with_options` follows the same pattern. `internal_scan_with_metrics` is removed; its metrics logic is inlined into both callers.

Supporting-rule filtering semantics are preserved: supporting matches are kept through validation (for match pairing / template variable resolution) and dropped only after `validate_matches` completes.

## Testing

- All 435 existing tests pass.
- New regression test `test_parallel_scan_with_validate_does_not_panic`: spawns 64 threads each calling `scan_with_options(validate=true)` concurrently — this panicked before the fix, passes after.

## Performance

No impact. The same operations execute in the same order; no extra allocations or data passes are introduced. `validate_matches` (rayon work) now runs outside the `block_on` window, which is marginally better.

### DIFF
--- a/sds/src/scanner/mod.rs
+++ b/sds/src/scanner/mod.rs
@@ -493,7 +493,22 @@ impl Scanner {
         event: &mut E,
         options: ScanOptions,
     ) -> Result<Vec<RuleMatch>, ScannerError> {
-        block_on(self.internal_scan_with_metrics(event, options))
+        let start = Instant::now();
+        let validate = options.validate_matches;
+        // Collect matches inside block_on, then run finalize_matches (which uses rayon) outside of
+        // it to avoid re-entrancy between the futures LocalPool executor and RAYON_THREAD_POOL.
+        let result = block_on(self.internal_scan_collect(event, options));
+        match result {
+            Ok((mut rule_matches, io_duration)) => {
+                self.finalize_matches(&mut rule_matches, validate);
+                self.record_metrics(&rule_matches, start, Some(io_duration));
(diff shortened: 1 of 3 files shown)


fuchsnj (approved):
This seems like a reasonable workaround. There's probably some fundamental async architecture changes that could be made to clean this up a bit / make this not necessary but that would be a bigger overhaul.

PR #357, opened 2026-05-13
Refactor into a single crate

fuchsnj:
This refactors the repo into a single Rust crate so make publishing to crates.io easier. Each crate is now part of the single `dd-sensitive-data-scanner` crate but behind separate feature flags so you can still pull in only the parts you want.

This is just changing the code organization around, there is no new features, except the new CI job for publishing the crate.


isabella-garza-datadog (approved):
LGTM overall!

One question I had while reviewing this was how it impacts integrations that build `dd-sds` directly. I took a look at agentless, and it seems like they build the `sds-go` crate directly ([link](https://github.com/DataDog/datadog-agentless-scanner/blob/cb152e0aafa95dc5bcebbbcd965929273c318f47/scripts/build.sh#L67)). If I’m understanding correctly, this would be a breaking change for that setup and would require updating how SDS is built in that repo, right? It’s probably worth creating a ticket for the Data Security pod to track that follow-up.

Also curious if you know of any other integrations that might be impacted by this change.

fuchsnj:
> One question I had while reviewing this was how it impacts integrations that build dd-sds directly. I took a look at agentless

That is using the `sds-go` Go bindings, which should not be impacted by this change. Only the Rust crates were merged, but the go bindings should be the same.

PR #358, opened 2026-05-22
Auto-run publish job

fuchsnj:
[Jira](https://datadoghq.atlassian.net/browse/SDSP-138)

The crates.io publish job works, so I'm enabling it on merges to `main` automatically.

### DIFF
--- a/.github/workflows/publish-crate.yml
+++ b/.github/workflows/publish-crate.yml
@@ -65,12 +65,12 @@ jobs:
         run: cargo package --manifest-path="sds/Cargo.toml" --allow-dirty
 
       - name: Authenticate with crates.io
-        if: ${{ !inputs.dry_run || inputs.dry_run == 'false' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-crate.yml
```diff
@@ -62,12 +65,12 @@ jobs:
         run: cargo package --manifest-path="sds/Cargo.toml" --allow-dirty
 
       - name: Authenticate with crates.io
-        if: ${{ !inputs.dry_run }}
+        if: ${{ !inputs.dry_run || inputs.dry_run == 'false' }}
```
isabella-garza-datadog: nit: Since `inputs` are unset on `push` events, could we make the intended behavior explicit? For example, auth/publish could use `if: ${{ github.event_name == 'push' || !inputs.dry_run }}` to say “publish on main pushes, and publish manual runs unless dry_run is true.”

isabella-garza-datadog (approved):
LGTM, left one small comment

PR #367, opened 2026-07-13
[DSEC-112] add opt-out feature for third-party active checkers

aimenebelfodil:
## Motivation

The Datadog Agent will ship the SDS scanner by **importing `dd-sds` directly as a Rust crate dependency**. In that use case the third-party active checkers are **not** used, so we want to avoid pulling in the large dependency set they require (HTTP client, full TLS stack, AWS request signing, rayon) when it isn't needed. This PR adds a way to compile the scanner without those checkers and their dependencies, while keeping them enabled by default for everyone else.

## Summary

Adds a default-on `third-party-active-checkers` Cargo feature so the scanner can optionally be built **without** third-party active checkers (HTTP/AWS match validation) and their exclusive dependencies. Behavior is unchanged by default.

The three dependencies used *exclusively* by third-party checkers are made optional and gated behind the feature:
- `reqwest` (+ its HTTP/TLS stack)
- `aws-sign-v4`
- `rayon` (only used to parallelize match validation)

Shared deps (`tokio`, `futures`, `num_cpus`, `moka`, `chrono`) are untouched since they're needed by the core scanning/async paths.

Config/status types (`MatchValidationType`, `MatchStatus`, etc.) and the scanner integration stay always-compiled, so the **public API and Go bindings are identical** in both configurations — only the network-backed validator code and its deps are gated.

## How to import without third-party checkers (Rust consumers, e.g. the Agent)

```toml
[dependencies]
dd-sensitive-data-scanner = { git = "https://github.com/DataDog/dd-sensitive-data-scanner", default-features = false, features = ["dd-sds"] }
```

(`default-features = false` drops the checkers; `dd-sds` is required since the whole public API is gated behind it.)

Build commands:

```bash
# Rust core only
cargo build --manifest-path=sds/Cargo.toml --release --no-default-features --features dd-sds
# Go bindings without checkers
make build-sds-go-no-third-party-active-checkers
```

## Impact

The Agent imports `dd-sds` as a Rust crate and links the `rlib`, so the relevant metric is the compiled Rust footprint. Disabling the checkers removes ~79 transitive crates (215 → 136 in `cargo tree`; 201 → 127 built `rlib`s), dropping the reqwest/hyper/rustls/ring/aws-sign/rayon stack and linking no outbound-network/TLS code.

### rlib

| Artifact (release) | With checkers (default) | Without checkers | Delta |
|---|---|---|---|
| `libdd_sds.rlib` (dd-sds's own code) | 5.77 MiB | 5.14 MiB | −0.63 MiB |
| Total compiled `rlib`s (dd-sds + all deps) | 204.6 MiB | 142.1 MiB | −62.5 MiB (−31%) |

dd-sds's own `rlib` barely changes (only the validator code is gated out); the real win is the **~62 MiB / 31% reduction in transitive dependency code**, plus lower compile time and supply-chain/audit surface. (Final linked-binary savings are smaller after dead-code elimination, but the dependency graph and build cost drop substantially.)

### dylib / staticlib (Go bindings artifacts, for reference)

| Build | Third-party checkers | `libdd_sds.dylib` |
|---|---|---|
| with `dd_sds_go` (Go bindings) | on (default) | 9.12 MiB |
| with `dd_sds_go` (Go bindings) | off | 4.52 MiB |
| without `dd_sds_go` (`dd-sds` core only) | on (default) | 0.40 MiB |
| without `dd_sds_go` (`dd-sds` core only) | off | 0.35 MiB |

For the Go bindings the `dylib` is roughly **halved (~4.6 MiB saved)**. (The non-Go `dd-sds` cdylib is tiny in both cases because it exports no C symbols without the Go bindings, so the linker dead-strips almost everything — it isn't a meaningful deliverable on its own; a Rust consumer links the `rlib`.)

## Changes
- `Cargo.toml`: make `reqwest`/`aws-sign-v4`/`rayon` optional; add default-on `third-party-active-checkers` feature.
- Gate validator modules, the `reqwest`-based code, `RAYON_THREAD_POOL`, `validate_matches`, and validator creation behind the feature.
- `Makefile`/`rust_checks.sh`: add feature-off build + clippy checks; new no-checkers Go build target.
- `generate_license_3rdparty.sh`: update hardcoded default-feature line.

## Test plan
- [x] `cargo clippy -- -D warnings` passes with feature on (`dd_sds_go`) and off (`--no-default-features --features dd-sds,dd_sds_go`)
- [x] `cargo tree` confirms `reqwest`/`aws-sign-v4`/`rayon` dropped when the feature is off
- [x] All 94 match-validation tests pass on the default build
- [x] `make check-rust` passes
- [x] CI green

### DIFF
--- a/scripts/rust_checks.sh
+++ b/scripts/rust_checks.sh
@@ -4,6 +4,8 @@ cargo check --manifest-path="sds/Cargo.toml" --benches --features dd_sds_go
 cargo check --manifest-path="sds/Cargo.toml" --bin fuzz --features sds-fuzz
 cargo clippy --manifest-path="sds/Cargo.toml" --features dd_sds_go -- -D warnings
 cargo clippy --manifest-path="sds/Cargo.toml" --bin fuzz --features sds-fuzz -- -D warnings
+# Ensure the crate also builds cleanly without third-party active checkers (no reqwest/aws-sign-v4).
+cargo clippy --manifest-path="sds/Cargo.toml" --no-default-features --features dd-sds,dd_sds_go -- -D warnings
 
 DID_STASH=0
 
--- a/Makefile
+++ b/Makefile
@@ -13,9 +13,14 @@ help:     ## Show this help.
 ##@ Build sds-go
 
 .PHONY: build-sds-go
-build-sds-go: ## Build the sds-go lib.
+build-sds-go: ## Build the sds-go lib (third-party active checkers included by default).
 	@echo "Building sds-go lib"
 	cargo build --manifest-path="sds/Cargo.toml" --release --features dd_sds_go
+
+.PHONY: build-sds-go-no-third-party-active-checkers
+build-sds-go-no-third-party-active-checkers: ## Build the sds-go lib without third-party active checkers (drops reqwest/aws-sign-v4).
+	@echo "Building sds-go lib without third-party active checkers"
+	cargo build --manifest-path="sds/Cargo.toml" --release --no-default-features --features dd-sds,dd_sds_go
 
 ##@ Formatting
 
@@ -45,6 +50,8 @@ test-rust: ## Test the rust lib.
 	@echo "Testing rust lib"
 	cargo test --manifest-path="sds/Cargo.toml"
 	cargo test --manifest-path="sds/Cargo.toml" --features dd_sds_go
+	@echo "Checking rust lib builds without third-party active checkers"
+	cargo check --manifest-path="sds/Cargo.toml" --no-default-features --features dd-sds,dd_sds_go
 
 .PHONY: test-all
 test-all: test-rust test-go ## Test the rust lib and golang libs.
--- a/scripts/generate_license_3rdparty.sh
+++ b/scripts/generate_license_3rdparty.sh
@@ -30,8 +30,8 @@ import sys
 
 path = pathlib.Path(sys.argv[1])
 contents = path.read_text()
-default_features = 'default = ["dd-sds"]'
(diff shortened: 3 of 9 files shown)


scripts/rust_checks.sh
```diff
@@ -4,6 +4,8 @@ cargo check --manifest-path="sds/Cargo.toml" --benches --features dd_sds_go
 cargo check --manifest-path="sds/Cargo.toml" --bin fuzz --features sds-fuzz
 cargo clippy --manifest-path="sds/Cargo.toml" --features dd_sds_go -- -D warnings
 cargo clippy --manifest-path="sds/Cargo.toml" --bin fuzz --features sds-fuzz -- -D warnings
+# Ensure the crate also builds cleanly without third-party active checkers (no reqwest/aws-sign-v4).
+cargo clippy --manifest-path="sds/Cargo.toml" --no-default-features --features dd-sds,dd_sds_go -- -D warnings
```
fuchsnj, line 8: This is fine for now, we should probably switch over to using `cargo hack` at some point to automate this. I'll make a note to myself to clean this up in the future. (for the whole repo, not just your changes)

fuchsnj (approved):

scripts/rust_checks.sh
fuchsnj: https://datadoghq.atlassian.net/browse/SDSP-488

aimenebelfodil: Sounds good for me thanks @fuchsnj !

PR #369, opened 2026-07-20
Update the proximity_keyword logic to accept the absence of a separator

Taiki-San:
Address issues where a rule looking to match `access key` wouldn't match `accessKey`

### DIFF
--- a/sds/src/proximity_keywords/included_keywords.rs
+++ b/sds/src/proximity_keywords/included_keywords.rs
@@ -280,5 +280,16 @@ mod test {
             vec![0]
         );
         assert!(collect_keyword_matches(keywords.keyword_matches("host ping")).is_empty());
+
+        // Same reasoning for a trailing link char: the "-" in "host-" must stay mandatory,
+        // so it requires an actual link char after "host" and won't match bare "host" with
+        // nothing following it.
+        let keywords = compile_keywords(30, &["host-"]);
+
+        assert_eq!(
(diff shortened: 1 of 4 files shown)


sds/src/proximity_keywords/included_keywords.rs
```diff
@@ -246,4 +246,39 @@ mod test {
...
+            collect_keyword_matches(keywords.keyword_matches("bla.ble secret")),
+            vec![0]
+        );
+    }
+
+    #[test]
+    fn leading_and_trailing_link_chars_stay_mandatory() {
+        // The leading "-" is part of the keyword's own shape, not a word separator, so it
+        // must not become optional (otherwise "-host" would also match bare "host").
+        let keywords = compile_keywords(30, &["-host"]);
```
isabella-garza-datadog, line 276: can you add a similar test case for trailing link chars?

sds/src/proximity_keywords/included_keywords.rs
```diff
@@ -246,4 +246,39 @@ mod test {
             vec![0]
         );
     }
+
```
isabella-garza-datadog, line 249: This change would also apply to excluded keywords. Can you add a minimal test for this in `excluded_keywords.rs`?

isabella-garza-datadog (commented):
Overall seems like a good approach. My comments are focused on adding unit test coverage.

isabella-garza-datadog (approved):

PR #373, opened 2026-07-24
Simplify CI

fuchsnj:
[Jira](https://datadoghq.atlassian.net/browse/SDSP-488)

This does some general CI cleanup. Mostly switching out some of the Rust feature testing for `cargo hack` to automatically test all features individually

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ format-all: format-rust format-go ## Format the rust lib and golang libs.
 .PHONY: test-go
 test-go: ## Test the golang lib.
 	@echo "Testing golang lib"
-	cd sds-go/go && go test ./...
+	cd sds-go/go && LD_LIBRARY_PATH="$(ROOT_DIR)/sds/target/release" DYLD_LIBRARY_PATH="$(ROOT_DIR)/sds/target/release" go test ./...
 	
 .PHONY: test-rust
 test-rust: ## Test the rust lib.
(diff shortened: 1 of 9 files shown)


scripts/rust_checks.sh
```diff
@@ -1,27 +1,10 @@
...
-if ! git diff --quiet || ! git diff --cached --quiet; then
-    git stash push -u -m "rust_checks.sh temp stash" >/dev/null 2>&1 || true
-    DID_STASH=1
-fi
-
-cargo fmt --check --manifest-path="sds/Cargo.toml" --all
-git diff --exit-code
\ No newline at end of file
+cargo hack check --each-feature --no-dev-deps
+cargo hack clippy --each-feature --no-dev-deps -- -D warnings
```
artslidd, line 8: curious how this works?

Makefile
```diff
@@ -15,12 +15,12 @@ help:     ## Show this help.
 .PHONY: build-sds-go
 build-sds-go: ## Build the sds-go lib (third-party active checkers included by default).
 	@echo "Building sds-go lib"
-	cargo build --manifest-path="sds/Cargo.toml" --release --features dd_sds_go
+	cd sds && cargo build --release --features dd_sds_go
```
artslidd, line 18: how is `cd ... &&` better than using the manifest-path command arg?

scripts/rust_checks.sh
fuchsnj: This basically just does a loop over all of the features and builds the library with each individual one. This will help prevent us from having to manually list the different features anytime it changes.

Makefile
fuchsnj: It's not really. Just slightly shorter / easier to read. Should do the same thing

scripts/rust_checks.sh
artslidd: nice that's neat!

Makefile
artslidd: ok ok :)

artslidd (approved):
LGTM

PR #374, opened 2026-07-24
Generate and verify Go FFI bindings

fuchsnj:
[Jira](https://datadoghq.atlassian.net/browse/SDSP-488)

The checked-in C header can drift from the Rust FFI while the Go bindings still compile, leaving runtime failures for downstream consumers. This generates the header from Rust with cbindgen and adds one CI check that verifies the header, builds the shared library, and runs the uncached Go binding suite.

This should help prevent issues such as ones fixed by [this PR](https://github.com/DataDog/dd-sensitive-data-scanner/pull/372)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,4 +11,18 @@ This contains:
 - An extensive library of secondary validators / checksums
 - Go bindings
 
+## Updating the Go bindings
+
+The Go bindings use the generated C header at `sds-go/go/dd_sds.h`. When an
+exported Rust function in `sds/src/native/` is added, removed, or its signature
+changes, install `cbindgen` 0.29.2 and regenerate the header:
+
+```shell
+cargo install cbindgen --version 0.29.2 --locked
+make update-sds-go-header
+```
+
+Commit the regenerated header with the Rust and Go changes. CI runs
+`make check-sds-go-bindings` to verify that the header is current and that the
+Go bindings build and pass their tests against the Rust library.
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -11,4 +11,18 @@ This contains:
...
 - Go bindings
 
+## Updating the Go bindings
+
+The Go bindings use the generated C header at `sds-go/go/dd_sds.h`. When an
+exported Rust function in `sds/src/native/` is added, removed, or its signature
+changes, install `cbindgen` 0.29.2 and regenerate the header:
+
+```shell
+cargo install cbindgen --version 0.29.2 --locked
```
vinckama, line 21: To make this need self documented in the command if it's missing, would it be possible to add a pre-check before re-generating the bindings:
```
.PHONY: check-cbindgen
check-cbindgen: ## Ensure cbindgen is installed for Go binding header generation.
	@if ! command -v cbindgen >/dev/null 2>&1; then \
		echo "error: cbindgen is not installed (required for Go binding header generation)." >&2; \
		echo "Install with: cargo install cbindgen --version 0.29.2" >&2; \
		exit 1; \
	fi

...
```
and then calling the check in `update-sds-go-header`
```
update-sds-go-header: check-cbindgen

vinckama (commented):
I added a suggestion

README.md
fuchsnj: fixed

vinckama (approved):

PR #376, opened 2026-08-06
[DSEC-232] Provide reset_regex_caches

aimenebelfodil:
## Motivation

Regex scratch caches (the lazy-DFA buffers, tens of KB per rule) are pooled globally with one slot per CPU and live for the process lifetime. Each thread that scans lazily fills its slot with a cache per rule, so scratch memory scales with `rules × threads` and stays resident even when idle.

For episodic-scanning embedders (e.g. the Agent's datasecurity check) that build a scanner, scan, then drop it, there was no way to reclaim this: dropping the scanner leaves the pooled scratch behind, and a per-thread GC can only reach the calling thread's slot — not caches populated by other threads.

## Change

Add `reset_regex_caches()`: swaps the global pool for a fresh one under an `RwLock`, dropping every thread's cached scratch at once. Scans clone the pool `Arc` under a short read lock, so an in-flight scan keeps the old pool alive until it finishes — safe to call concurrently. Caches are lazily recreated on the next scan.

Call it when scanning is idle to return the scanner's scratch footprint to ~0, regardless of how many threads scanned.

### DIFF
--- a/sds/src/lib.rs
+++ b/sds/src/lib.rs
@@ -110,7 +110,10 @@ pub use scanner::{
         ClaimRequirement, JwtClaimsValidatorConfig, ProximityKeywordsConfig, RegexRuleConfig,
         SecondaryValidator,
     },
-    regex_rule::{RegexCacheKey, RegexCacheValue, RegexCaches, SharedRegex, get_memoized_regex},
+    regex_rule::{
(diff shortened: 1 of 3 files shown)


vhourdel (approved):
LGTM if you're sure about the unwraps

PR #377, opened 2026-08-06
[DSEC-232] Provide clear_all_caches

aimenebelfodil:
## Motivation

Stacked on #376 (`reset_regex_caches`). Adds the second half so episodic embedders (build → scan → drop) can reclaim everything when idle.

- `gc_regex_store()` — drops compiled-regex store entries no longer referenced by any live scanner (otherwise only pruned every `GC_FREQUENCY` inserts).
- `clear_all_caches()` — resets the per-thread scratch pool and GCs the regex store in one call.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/sds/src/lib.rs
+++ b/sds/src/lib.rs
@@ -111,8 +111,8 @@ pub use scanner::{
         SecondaryValidator,
     },
     regex_rule::{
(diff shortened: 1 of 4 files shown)


vhourdel (approved):
LGTM

PR #381, opened 2026-08-20
sds-go: add Suppressions to RegexRuleConfig

vinckama:
The Rust engine already compiles starts_with / ends_with / exact_match lists and drops matching hits, but the Go bindings never sent that field over FFI.  Go callers therefore could not suppress matches through this package.

Add the same value-type Suppressions model used in sds-shared-library, thread it through ExtraConfig and the New*Rule constructors, and map CreateScannerError::InvalidSuppressions (-6) to ErrInvalidSuppressions.

Needed for [DATASEC-204](https://datadoghq.atlassian.net/browse/DATASEC-204).

[DATASEC-204]: https://datadoghq.atlassian.net/browse/DATASEC-204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sds-go/go/scanner.go
+++ b/sds-go/go/scanner.go
@@ -22,6 +22,7 @@ var (
 	ErrInvalidRegex                 error = fmt.Errorf("invalid regex")
 	ErrInvalidKeywords              error = fmt.Errorf("invalid keywords")
 	ErrInvalidMatchAction           error = fmt.Errorf("invalid match action")
+	ErrInvalidSuppressions          error = fmt.Errorf("invalid suppressions")
 	ErrSupportingRuleHasMatchAction error = fmt.Errorf("supporting rules cannot have a match action other than None")
 )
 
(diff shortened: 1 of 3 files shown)


vinckama:
@codex review

sds-go/go/scanner.go
```diff
@@ -115,6 +116,8 @@ func CreateScannerWithOptions(ruleConfigs []RuleConfig, options ScannerOptions)
 			} else {
 				return nil, fmt.Errorf("internal panic")
 			}
+		case -6: // rust: CreateScannerError::InvalidSuppressions
```
vinckama, line 119: `case -7` is invalid capture group which is not supported in this PR, this case would be needed when capture group are supported

fuchsnj (approved):

PR #385, opened 2026-08-27
sds-go: unmarshal JwtClaimsValidator config

vinckama:
The Go bindings already constructed JwtClaimsValidator rules and sent them over FFI, but SecondaryValidator had no UnmarshalJSON, so loading a rule from JSON left Config as nested maps instead of JwtClaimsValidatorConfig.

For DATASEC-201

### DIFF
--- a/sds-go/go/regex_rule.go
+++ b/sds-go/go/regex_rule.go
@@ -123,6 +123,51 @@ func NewJwtClaimsValidator(config JwtClaimsValidatorConfig) *SecondaryValidator 
 	}
 }
 
+// UnmarshalJSON handles deserialization from various JSON formats
+func (v *SecondaryValidator) UnmarshalJSON(data []byte) error {
+	// Try to unmarshal as a simple string first (legacy format)
+	var str string
+	if err := json.Unmarshal(data, &str); err == nil {
+		v.Type = SecondaryValidatorType(str)
+		v.Config = nil
+		return nil
+	}
+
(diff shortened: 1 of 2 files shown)


vinckama:
@codex review

sds-go/go/regex_rule.go
```diff
@@ -123,6 +123,51 @@ func NewJwtClaimsValidator(config JwtClaimsValidatorConfig) *SecondaryValidator
...
+	}
+
+	v.Type = SecondaryValidatorType(rawMap.Type)
+
+	// Handle configuration if present
+	if len(rawMap.Config) > 0 {
+		switch rawMap.Type {
+		case string(JwtValidatorType):
+			var jwtConfig JwtClaimsValidatorConfig
+			if err := json.Unmarshal(rawMap.Config, &jwtConfig); err != nil {
```
vhourdel, line 152: This unmarshal doesn't seem to be supporting `NotExpired` (defined [here](https://github.com/DataDog/dd-sensitive-data-scanner/blob/22468ce5842d904a808c888c77d91d40b8c72796/sds/src/secondary_validation/jwt_claims_validator.rs#L111-L123) in the Rust engine). Maybe this should be tackled in a follow up PR.

vhourdel (approved):

sds-go/go/regex_rule.go
vinckama: good catch, I will add this in a follow-up PR.

PR #387, opened 2026-08-28
Improve match logging

Taiki-San:
The current logging emit a metric for every match, but with almost no metadata.
This makes fleet monitoring challenging since an increase in activity can't be tied to a precise rule, and the performance of a given rule can't easily be isolated.

This change introduce a new `sds_rule_name` tag with the category and name of the rule

### DIFF
--- a/sds/src/scanner/metrics.rs
+++ b/sds/src/scanner/metrics.rs
@@ -1,16 +1,23 @@
 use crate::Labels;
 use metrics::{Counter, counter};
-use std::collections::BTreeMap;
+
+/// Looks up the value of a `"key:value"` tag entry by key, matching the format used by
+/// standard rule definitions (e.g. `sensitive_data:travis_ci_access_token`).
+fn get_tag_value<'a>(tags: &'a [String], key: &str) -> Option<&'a str> {
+    let prefix_len = key.len() + 1;
+    tags.iter().find_map(|tag| {
+        (tag.len() > prefix_len && tag.starts_with(key) && tag.as_bytes()[key.len()] == b':')
+            .then(|| &tag[prefix_len..])
+    })
+}
 
 /// Computes the `"{sensitive_data_category}/{sensitive_data}"` tag value for a rule from its
 /// `tags`. Missing `sensitive_data_category` falls back to `"missing_category"`. Missing
 /// `sensitive_data` yields `None`, since there's nothing meaningful to tag with.
-pub fn compute_sds_rule_name(tags: &BTreeMap<String, String>) -> Option<String> {
-    let sensitive_data = tags.get("sensitive_data")?;
-    let sensitive_data_category = tags
-        .get("sensitive_data_category")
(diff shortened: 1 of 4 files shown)


sds/src/scanner/mod.rs
```diff
@@ -209,6 +222,10 @@ pub struct RootCompiledRule {
     pub suppressions: Option<CompiledSuppressions>,
     pub precedence: Precedence,
     pub is_supporting_rule: bool,
+    /// Precomputed `"{sensitive_data_category}/{sensitive_data}"` tag value derived from
+    /// the rule's `tags`, used to tag `scanning.match_count`. `None` when the rule has no
+    /// `sensitive_data` tag.
+    pub sds_rule_name: Option<String>,
```
vinckama, line 228: could it be kept generic with "labels" similarly to how these telemetry tags are attached to [scanner](https://github.com/DataDog/dd-sensitive-data-scanner/blob/9c648800a136cb730947fe74d68d7d13a5d77f3d/sds/src/scanner/mod.rs#L504)?

sds/src/scanner/mod.rs
```diff
@@ -585,10 +602,26 @@ impl Scanner {
...
+                .or_default() += 1;
+        }
+        for (rule_index, count) in match_counts_by_rule {
+            match self.rules[rule_index].sds_rule_name.as_deref() {
+                Some(sds_rule_name) => {
+                    let labels = self.labels.clone_with_labels(Labels::new(&[(
+                        "sds_rule_name",
+                        sds_rule_name.to_string(),
+                    )]));
+                    counter!("scanning.match_count", labels).increment(count);
```
vinckama, line 620: The [cardinality of this metric is already high](https://app.datadoghq.com/metric/summary?metric=libsds.scanning.match_count) and it will increase even more with this new tag. 

To avoid cardinality issues, these additional labels must be enabled as opt-in when required, for instance for correctness validation.  

If `labels` are forwarded when the rule is instantiated, the business logic to decide if they should be enabled or not could be done outside of the library. 
Such logic could be defined in the SDS context, enabled per target and propagated to downstream consumers.

sds/src/scanner/metrics.rs
```diff
@@ -1,6 +1,26 @@
 use crate::Labels;
 use metrics::{Counter, counter};
 
+/// Looks up the value of a `"key:value"` tag entry by key, matching the format used by
+/// standard rule definitions (e.g. `sensitive_data:travis_ci_access_token`).
```
vinckama, line 5: Minor: there is no concept of standard rules here, and in general the concept of `sensitive_data` tag is not associated to standard rule but is generic for all kind of rules.

vinckama (commented):
I added a few comments on the approach. I think the labels could stay generic and be associated to the rule when created. The business logic to select the label can be done upstream. This will allow to limit the impact on the metric cardinality.

PR #391, opened 2026-09-18
Add VIN checksum secondary validator

davidioi:
Adds the offline `VinChecksum` secondary validator so VIN-shaped matches with an incorrect ninth-position check digit can be rejected before redaction. For example, `L5BGA2V58NG590409` is rejected, while `LFWADRJF011002346` passes.

The validator implements the positional checksum in NHTSA 49 CFR 565.15(c) and GB 16735-2019 Appendix A. It requires 17 ASCII VIN characters, excludes I/O/Q, and normalizes ASCII case. It validates checksum integrity without claiming a vehicle was issued. The existing serialized validator interface is extended with `{"type":"VinChecksum"}`. No rule configuration is activated by this change; consumers must deploy support before accepting configurations that reference the new validator.

To clear the repository-required dependency check, this also updates only the lockfile entries for `rustls` 0.23.40 → 0.23.45 and `rustls-webpki` 0.103.13 → 0.103.15. `make check-deny` reproduced [RUSTSEC-2026-0285](https://rustsec.org/advisories/RUSTSEC-2026-0285.html) on the unchanged base and passed with this minimal update. No dependencies were added, and no advisory was suppressed.

### Validation

- The seven new Rust behavioral tests failed on the unchanged engine because the serialized validator was unsupported, then passed after implementation.
- `make check-rust`: passed, including the feature matrix, Clippy, all-target/all-feature checks, and formatting with warnings denied.
- Final locked all-feature Rust tests: 460 passed, zero failures or ignored tests.
- Fresh release Go native build: passed. Full Go suite: 47 top-level tests plus 28 subtests passed; zero skipped, including all nine VIN FFI cases.
- Go staticcheck, formatting, and example build: passed.
- `make check-licenses`: passed; the existing license inventory remains unchanged.
- `make check-deny` with cargo-deny 0.19.8: passed all categories.
- Locked all-feature dependency metadata resolution: passed.
- Published head `1e83807b3af2458a968ca9bf44f20c694073c05c` exactly matches the locally validated Git tree `8c4f2a318e6a30bbee79850ef66d665cbaa47494`; all five commits have GitHub-verified signatures.

Tests cover independent checksum vectors, every allowed transliteration character, malformed and non-ASCII input, case handling, all possible check-position characters, serialization, and scanning past an invalid candidate to a valid one.

An additional non-required `cargo clippy --all-targets -- -D warnings` invocation reports 13 existing test-code warnings on both the unchanged base and the VIN implementation. The repository-required `make check-rust` target passes; no unrelated test code was modified.

These are local validation results. GitHub CI and required maintainer review remain separate merge gates.


sds/Cargo.lock
```diff
@@ -2656,9 +2656,9 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.40"
+version = "0.23.45"
```
aimenebelfodil: Do you think we can split this part from the PR @davidioi  so the PR remains targeting only Vin Checksum 🙇

aimenebelfodil (commented):
Requested a small split of the PR but the VIN checksum seems correct!

aimenebelfodil (approved):
LGTM, I think we can make the PR description leaner a bit and strip out unnecessary details: typically this sentence 
**An additional non-required cargo clippy --all-targets -- -D warnings invocation reports 13 existing test-code warnings on both the unchanged base and the VIN implementation. The repository-required make check-rust target passes; no unrelated test code was modified.** is outside the scope of the PR description

That's said, probably we should create a JIRA tasks regarding these tests warning as a follow up

sds/src/secondary_validation/vin_checksum.rs
```diff
@@ -0,0 +1,38 @@
...
+            sum += value * weight;
+        }
+
+        let check_digit = match sum % 11 {
+            10 => b'X',
+            remainder => b'0' + remainder as u8,
+        };
+        bytes[8].to_ascii_uppercase() == check_digit
+    }
+}
```
alexandrefouchs: @davidioi why add a test vin_checksum_test.go‎ and test/vin_checksum.rs and not a #[cfg(test)] section here like other validators?

